PR #17, opened 2020-01-10
Need user input pr field + Sorting

janid1967:


janid1967:
@hbjorgo Can You approve?

janid1967:
> Some tests have the postfix "Test", other not. Please add to those that don't have it.

No, We agreed to something else

PR #31, opened 2020-01-16
Move next due date column

janid1967:
Moved NextDueTimeUtc column from PreservationRecord to Requirement.
New method on Requirement: SetNextDueTimeUtc
PreservationRecord.PreservedBy renamed to PreservedByPersonId 
Nullable columns in PreservationRecord made not-nullable, and set via ctor.
Updated unit tests

### DIFF
--- a/src/Equinor.Procosys.Preservation.Domain/AggregateModels/TagAggregate/Requirement.cs
+++ b/src/Equinor.Procosys.Preservation.Domain/AggregateModels/TagAggregate/Requirement.cs
@@ -26,6 +26,7 @@ namespace Equinor.Procosys.Preservation.Domain.AggregateModels.TagAggregate
         }
 
         public int IntervalWeeks { get; private set; }
+        public DateTime? NextDueTimeUtc { get; private set; }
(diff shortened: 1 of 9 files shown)


src/Equinor.Procosys.Preservation.Domain/AggregateModels/TagAggregate/Requirement.cs
```diff
@@ -42,5 +43,15 @@ public void AddPreservationRecord(PreservationRecord preservationRecord)
 
             _preservationRecords.Add(preservationRecord);
         }
+
+        public void SetNextDueTimeUtc(DateTime currentTime)
```
hbjorgo: Wouldn't it be better to set this when a new preservationRecord is added? That method would take the PreservedAt time from that record, add the interval and update the NextDueTime property.

janid1967: Agree. Rename method AddPreservationRecord to Preserve

PR #34, opened 2020-01-20
Moved NeedUserInput

janid1967:
Moved NeedUserInput from Dto to Domain Objects (Field and RequirementDefinition). 
Need this info pr. Requirement when handling commands (StartPreservation, Preserve, etc.)
Unit tests updated


src/Equinor.Procosys.Preservation.Domain/AggregateModels/RequirementTypeAggregate/Field.cs
```diff
@@ -40,6 +40,10 @@ protected Field() : base(null)
         public bool? ShowPrevious { get; private set; }
         public int SortKey { get; private set; }
         public FieldType FieldType { get; private set; }
+        public bool NeedUserInput =>
```
hbjorgo: Consider renaming to "NeedsUserInput" in all places where that name is used.

PR #39, opened 2020-01-21
Unit test on all command handlers

janid1967:
Unit test on all missing CommandHandlers.
Abstract CommandHandlerTestsBase.cs is base on all test classes testing command handlers since IUnitOfWork and IPlantProvider is often needed


src/Equinor.Procosys.Preservation.Domain/AggregateModels/JourneyAggregate/Journey.cs
```diff
@@ -25,7 +25,7 @@ public Journey(string schema, string title)
         public void Void() => IsVoided = true;
         public void UnVoid() => IsVoided = false;
 
-        public void AddStep(Step step)
+        public virtual void AddStep(Step step)
```
hbjorgo: Is it possible to test without making this virtual?

janid1967: Yes.done

PR #45, opened 2020-01-23
CORS - allowing any header.

hbjorgo:


janid1967:
Nice if You can set Christer as reviewer on these types of changes.

PR #46, opened 2020-01-23
Project model

janid1967:
Refactor MainApiTagService.GetTagDetails to use endpoint GET /TagByNo
Project Model. Project is now AggregateRoot. Project contains Tags (Tag is not AggregateRoot)
Deleted column Tag.ProjectName since ProjectName is owned by parent Project
Unit tests for ProjectRepo, testing all methods

**NB Migration will fail if Tags exist due to constraint from Tag.ProjectId to new Project table. Tag table should be emptied before migrated**


src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/PreservationStatus.cs
```diff
@@ -1,7 +1,8 @@
-﻿namespace Equinor.Procosys.Preservation.Domain.AggregateModels.TagAggregate
+﻿namespace Equinor.Procosys.Preservation.Domain.AggregateModels.ProjectAggregate
 {
     public enum PreservationStatus
     {
+        // ReSharper disable once InconsistentNaming
```
hbjorgo: Either fix the code, update the editorconfig file or make sure that Resharper is using the editorconfig file for the project to avoid having tool-specific stuff in the code.

src/Equinor.Procosys.Preservation.MainApi/Tag/ProcosysTagDetails.cs
```diff
@@ -11,7 +11,9 @@ public class ProcosysTagDetails
         public string Description { get; set; }
         public string DisciplineCode { get; set; }
         public string McPkgNo { get; set; }
+        public string ProjectDescription { get; set; }
```
hbjorgo: Are these extra properties needed?

janid1967: Yes. Used to populate Project table when creating such

PR #54, opened 2020-01-29
Current user

hbjorgo:
Added functionality to get the current user.


janid1967:
Who is responsible to add current user into Person table?

PR #56, opened 2020-01-29
Enrich create tag command

janid1967:


src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/CreateTagDto.cs
```diff
@@ -4,9 +4,10 @@ namespace Equinor.Procosys.Preservation.WebApi.Controllers.Tags
 {
     public class CreateTagDto
     {
-        public string TagNo { get; set; }
+        public IEnumerable<string> TagNos { get; set; }
         public string ProjectName { get; set; }
         public int StepId { get; set; }
         public IEnumerable<TagRequirementDto> Requirements { get; set; }
+        public string Remarks { get; set; }
```
hbjorgo: Use singular since is is only one string?

janid1967: fixed

PR #61, opened 2020-01-30
FluentValidation validators in Swagger

hbjorgo:
Added library to read FluentValidation validators and add the rules to Swagger.


janid1967:
> To me it seems that it don't do what we want.
> Sample: CreateJourneyDtoValidator has .NotNull for Title. Swagger say nullable = true

Forget it ... I read the doc and noticed the red star beside of prop

PR #63, opened 2020-01-31
RequirementDTO.NextDueWeeks 

janid1967:
calculate NextDueWeeks so the number is the same during the week. I.e NextDueWeeks shifts at Monday night regardless of where in week the duedate / current date is


src/Equinor.Procosys.Preservation.Domain/DateTimeExtensions.cs
```diff
@@ -1,14 +1,45 @@
...
+
+        public static DateTime StartOfWeek(this DateTime dt)
+        {
+            var dayOfWeek = GetDayOfWeekMondayAsFirst(dt);
+            var diff = -(dayOfWeek - (int)firstDayOfWeek);
+            return dt.AddDays(diff).Date;
+
+        }
+
+        public static int GetWeeksReferencedFromStartOfWeek(this DateTime toDateTime, DateTime fromDateTime)
```
hbjorgo: Consider replacing it with this method:
```
public static int GetWeeksDiff(DateTime from, DateTime to, CultureInfo culture)
        {
            Calendar cal = culture.Calendar;
            CalendarWeekRule calendarWeekRule = culture.DateTimeFormat.CalendarWeekRule;
            DayOfWeek firstDayOfWeek = culture.DateTimeFormat.FirstDayOfWeek;

            int fromWeek = cal.GetWeekOfYear(from, calendarWeekRule, firstDayOfWeek);
            int toWeek = cal.GetWeekOfYear(to, calendarWeekRule, firstDayOfWeek);

            return toWeek - fromWeek;
        }
```

It gets the week number for both datetimes and subtracs the weeknumbers to get the diff.

janid1967: It fails when it passes years. Suggest we fix it later. I renamed method to GetWeeksUntil and switched input param to / from

PR #66, opened 2020-02-07
71225 Bulk-preserve allowed only if due or over due

janid1967:
Moved logic regarding ReadyToBePreserved from Query to Domain.
Must have both ReadyToBePreserved and ReadyToBeBulkPreserved property pr. Requirement and pr Tag
@arildbjerkenes This is a breaking change on DTO

### DIFF
--- a/src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Requirement.cs
+++ b/src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Requirement.cs
@@ -53,14 +53,14 @@ namespace Equinor.Procosys.Preservation.Domain.AggregateModels.ProjectAggregate
             return timeUtc.GetWeeksUntil(NextDueTimeUtc.Value);
         }
         
-        public bool IsReadyToBeBulkPreserved(DateTime preservedAtUtc)
+        public bool IsReadyToBeBulkPreserved(DateTime currentTimeUtc)
         {
             if (!ReadyToBePreserved)
             {
                 return false;
             }
 
-            var nextDueInWeeks = GetNextDueInWeeks(preservedAtUtc);
+            var nextDueInWeeks = GetNextDueInWeeks(currentTimeUtc);
 
             return nextDueInWeeks.HasValue && nextDueInWeeks.Value <= 0;
         }
--- a/src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Requirement.cs
+++ b/src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Requirement.cs
@@ -42,7 +42,29 @@ namespace Equinor.Procosys.Preservation.Domain.AggregateModels.ProjectAggregate
         public void UnVoid() => IsVoided = false;
 
         public bool ReadyToBePreserved => PeriodReadyToBePreserved != null;
+
+        public int? GetNextDueInWeeks(DateTime timeUtc)
+        {
+            if (!NextDueTimeUtc.HasValue)
+            {
+                return null;
+            }
+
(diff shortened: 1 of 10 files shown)


src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Requirement.cs
```diff
@@ -42,7 +42,29 @@ public Requirement(string schema, int intervalWeeks, RequirementDefinition requi
...
         public bool ReadyToBePreserved => PeriodReadyToBePreserved != null;
+
+        public int? GetNextDueInWeeks(DateTime timeUtc)
+        {
+            if (!NextDueTimeUtc.HasValue)
+            {
+                return null;
+            }
+
+            return timeUtc.GetWeeksUntil(NextDueTimeUtc.Value);
```
hbjorgo: Perhaps "GetWeeksUntil" could be moved to this class as a private helper method if it isn't used anywhere else. It is quite domain specific.

src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Tag.cs
```diff
@@ -122,7 +122,14 @@ public Requirement FirstUpcomingRequirement
                 .FirstOrDefault();
 
         public bool ReadyToBePreserved
-            => Status == PreservationStatus.Active && FirstUpcomingRequirement.ReadyToBePreserved;
+            => Status == PreservationStatus.Active && 
+               FirstUpcomingRequirement != null && 
+               FirstUpcomingRequirement.ReadyToBePreserved;
+
+        public bool IsReadyToBeBulkPreserved(DateTime preservedAtUtc)
```
hbjorgo: Is it the time it was preserved at that should be passed in here, or is it current time? If the latter, consider renaming to "now".

src/tests/Equinor.Procosys.Preservation.Domain.Tests/AggregateModels/ProjectAggregate/RequirementTests.cs
```diff
@@ -76,9 +77,104 @@ public void StartPreservation_ShouldSetActivePeriod()
         }
 
         [TestMethod]
-        public void StartPreservation_ShouldThrowException_WhenPreservationAlreadyOngoing()
+        public void StartPreservation_ShouldNotSetReadyToBePreserved_WhenFieldNeedInput()
```
hbjorgo: Typo, missing 's': "StartPreservation_ShouldNotSetReadyToBePreserved_WhenFieldNeedsInput"

src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Requirement.cs
janid1967: > Perhaps "GetWeeksUntil" could be moved to this class as a private helper method if it isn't used anywhere else. It is quite domain specific.

Can U suggest how to unit test it if we make it private?

src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Tag.cs
janid1967: Will rename. But the same name is consistent used many places so it will be a "global" rename

src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Requirement.cs
hbjorgo: Good point.

janid1967: I leave it as is

PR #70, opened 2020-02-11
71225 business rule for allow bulk only if due or over due

janid1967:


src/Equinor.Procosys.Preservation.Command/TagCommands/Preserve/BulkPreserve/BulkPreserveCommand.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace Equinor.Procosys.Preservation.Command.TagCommands.Preserve.BulkPreserve
+{
+    public class BulkPreserveCommand : PreserveCommand
+    {
+        public BulkPreserveCommand(IEnumerable<int> tagIds) : base(tagIds)
```
hbjorgo: Do we gain anything having a base class? If this class should be covered by unit tests, they must be written specifically for this class anyway to verify that the base class is called correctly...
I personally prefer each command being independent.

src/Equinor.Procosys.Preservation.Command/Validators/Tag/TagValidator.cs
```diff
@@ -69,6 +70,17 @@ public bool ReadyToBePreserved(int tagId)
             return tag.ReadyToBePreserved;
         }
 
+        public bool ReadyToBeBulkPreserved(int tagId, DateTime preservedAtUtc)
+        {
+            var tag = _projectRepository.GetTagByTagIdAsync(tagId).Result;
+            if (tag == null)
```
hbjorgo: If the command handler expects all input to be valid, should the validator also expect the input to be valid? Should all tag validators expect that the tag existence being validated first?

src/Equinor.Procosys.Preservation.Command/TagCommands/Preserve/BulkPreserve/BulkPreserveCommand.cs
janid1967: Nahh , this is to be able to re-use BusinessValidato. To be able to BulkPreserve, all rules for "normal" Preserve must be fulfilled first. BulkPreserve has one rule in addition

src/Equinor.Procosys.Preservation.Command/Validators/Tag/TagValidator.cs
janid1967: "Impossible" to chain tests in BusinessValidators if we expect tag to exists when checking if it is Voided or ReadyToBe"Something" ...

src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Tag.cs
```diff
@@ -131,15 +127,27 @@ public bool IsReadyToBeBulkPreserved(DateTime currentTimeUtc)
...
 
-        public void Preserve(DateTime preservedAtUtc, Person preservedBy, bool bulkPreserved)
+        public void Preserve(DateTime preservedAtUtc, Person preservedBy)
         {
             if (!ReadyToBePreserved)
             {
                 throw new Exception($"{nameof(Tag)} {Id} is not ready to be preserved ");
             }
-            foreach (var requirement in Requirements.Where(r => r.ReadyToBePreserved))
+            foreach (var requirement in UpComingRequirements().Where(r => r.ReadyToBePreserved))
```
hbjorgo: This will preserve all future requirements that have a due date. Perhaps there needs to be a check on the due date?

janid1967: Agree. Will create a new Task on same BPI 71225 for that corner case

PR #75, opened 2020-02-12
71225 preserve on single tag

janid1967:
Refactor:

PreserveCommand now take single tag (definition of not-bulk-preservation)
BulkPreserveCommand still take list of tags (definition of bulk-preservation)
Validation of PreserveCommand and BulkPreserveCommand is now separated.
Method Requirement.IsReadyToBeBulkPreserved renamed to IsReadyAndDueToBePreserved

**New logic when preserving a Tag (both bulk and not bulk): Preserve all upcomming requirements until first requirement not due reached (break loop)**


src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Tag.cs
```diff
@@ -169,9 +146,28 @@ public void UpdateCommentForActivePeriod(string comment, RequirementDefinition r
...
+        private void Preserve(DateTime preservedAtUtc, Person preservedBy, bool bulkPreserved)
+        {
+            if (!IsReadyToBePreserved(preservedAtUtc))
+            {
+                throw new Exception($"{nameof(Tag)} {Id} is not ready to be preserved ");
+            }
+
+            foreach (var requirement in UpComingRequirements())
+            {
+                if (!requirement.IsReadyAndDueToBePreserved(preservedAtUtc))
```
hbjorgo: Perhaps this check could be in UpComingRequirements() ?

janid1967: No, then it will be different logic. 
Let us say we have 3 requirements due: R1, R2 and R3 in that order. R1 and R3 are both due and ready but R2 lack input. Current logic will preserve R1 and then stop.
If we move check into where-clause of UpComingRequirements(), it will preserve R1 and R3.

Probably both methods will work in real life

hbjorgo: It depends on the definition of "Upcoming Requirements". If it is just R1 in you example, I would move the for-loop to UpComingRequirements so that it will return R1.

PR #80, opened 2020-02-15
Feature/71263 record values unit tests

janid1967:
Moved businesslogic regarding recording values on active period on Requirement from Tag class to Requirement class. 
Got rid of class FieldValue used in command and use Dictionary instead

Moved all tests regarding recording values on Requirement from RecordValuesCommandHandlerTests to RequirementTests.

Added more tests for Requirement


src/tests/Equinor.Procosys.Preservation.Domain.Tests/AggregateModels/ProjectAggregate/RequirementTests.cs
```diff
@@ -351,10 +399,559 @@ public void Preserve_ShouldAddNewPreservationPeriodEachTime()
...
+            // Assert
+            Assert.IsNull(dut.GetCurrentFieldValue(_numberFieldMock.Object));
+        }
+
+        #endregion
+
+        #region GetPreviousValue
+
+        [TestMethod]
+        public void GetPreviousFieldValue_ShouldReturnValue_AfterRecordingAndPreserved()
```
janid1967: Method GetPreviousFieldValue_ShouldReturnValue_AfterRecordingAndPreserved will be replaced. Have a better version not commited yet

PR #87, opened 2020-02-19
71180-Action db model

janid1967:
Action
Tag has list of Actions
Migration of Action

Action will have list of a new Domain entiy: ActionComments. Next PR

PR #91, opened 2020-02-19
71180-Action comment db model

janid1967:
based on #87 

@hbjorgo Review this after reviewing #87 and #87 merged into Master


janid1967:
No use for Comments in Action after discuss with customer. Close this

PR #113, opened 2020-02-28
New input validation rule for Tag number suffix

janid1967:


src/tests/Equinor.Procosys.Preservation.WebApi.Tests/Controllers/Tags/CreateAreaTagDtoValidatorTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            Assert.IsTrue(result.IsValid);
+        }
+
+        [TestMethod]
+        public void Fail_WhenWhiteSpaceInSuffix()
+        {
+            var dut = new CreateAreaTagDtoValidator();
+
+            validCreateAreaTagDto.TagNoSuffix = "10 A";
```
hbjorgo: Create another variable "invalidaCreateAreaTagDto". It is confusing to that validCreateAreaTagDto results in IsValid false.

PR #118, opened 2020-03-05
71671 Prepare for sorting on next due date

janid1967:
Persist Tag.NextDueTimeUtc instead of calculate from Requirements on the fly. This because we need this column for sorting in Scope. Migration to insert correct NextDueTimeUtc in existing Tags.
Sorting will come in next PR.


src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Requirement.cs
```diff
@@ -186,7 +186,7 @@ public void SetModified(DateTime modifiedAtUtc, Person modifiedBy)
         private PreservationPeriod PeriodReadyToBePreserved
             => PreservationPeriods.SingleOrDefault(pp => pp.Status == PreservationPeriodStatus.ReadyToBePreserved);
 
-        private void AddNewPreservationPeriod(DateTime offsetTimeUtc)
+        private void PrepareNewPreservation(DateTime offsetTimeUtc)
```
hbjorgo: Should this be called when a requirement is added/removed or otherwise edited? If so, maybe tag needs to subscribe so a Requirement.Changed event?

src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Tag.cs
```diff
@@ -239,9 +247,14 @@ private void Preserve(DateTime preservedAtUtc, Person preservedBy, bool bulkPres
...
                 requirement.Preserve(preservedAtUtc, preservedBy, bulkPreserved);
             }
+        
+            UpdateNextDueTimeUtc();
         }
 
         private Requirement FirstUpcomingRequirement(DateTime currentTimeUtc)
             => GetUpComingRequirements(currentTimeUtc).FirstOrDefault();
+
+        private void UpdateNextDueTimeUtc()
```
hbjorgo: Should this be called when a requirement is added/removed too? If a requirement is edited, what then? If so, maybe tag needs to subscribe to a Requirement.Changed event?

janid1967: Yes, but we don't know the business logic yet. Unit tests cover how it is now, and new considerations must be done when such logic added

janid1967: I will add UpdateNextDueTimeUtc() into AddRequirement method

PR #121, opened 2020-03-06
Unit of work

hbjorgo:


src/tests/Equinor.Procosys.Preservation.Command.Tests/Validators/JourneyValidatorTests.cs
```diff
@@ -14,9 +14,9 @@ public class JourneyValidatorTests : ReadOnlyTestsBase
         private const string JourneyTitle = "Journey";
         private int _journeyId;
         
-        protected override void SetupNewDatabase(DbContextOptions<PreservationContext> dbContextOptions)
+        protected override void SetupNewDatabase(DbContextOptions<PreservationContext> _dbContextOptions)
```
janid1967: undo (dbContextOptions is correct here)

PR #125, opened 2020-03-12
71692 Impl of IsNew logic for tags in scope

janid1967:


src/Equinor.Procosys.Preservation.WebApi/appsettings.json
```diff
@@ -26,5 +26,8 @@
     "BaseAddress": "",
     "ApiVersion": "",
     "TagSearchPageSize": 1000
+  },
+  "ApiOptions": {
```
hbjorgo: Rename to "Tag": { "IsNewHours": 48 }
I would consider Api to be more WebApi settings. This is a domain setting for tag.

src/Equinor.Procosys.Preservation.Infrastructure/ApiOptions.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Equinor.Procosys.Preservation.Infrastructure
```
hbjorgo: This is a domain setting, so it should be handled by the domain namespace. Maybe it should be persisted in the database to have a single source of truth?

janid1967: Moved, but I think database is "shooting spurv with canon"

PR #131, opened 2020-03-18
72077 filter on storage area

janid1967:
The last filter


src/Equinor.Procosys.Preservation.Query/GetTags/GetTagsQueryHandler.cs
```diff
@@ -228,6 +229,8 @@ where EF.Property<int>(req, "TagId") == tag.Id && request.Filter.RequirementType
                        tag.McPkgNo.StartsWith(request.Filter.McPkgNoStartsWith)) &&
```
TordJoranger: indentation to make it easier to see what opperators "belong" together.
Personally I find it easier if the operator comes before the condition 
(A == B
    || B == A)
&& (C ==D 
    || D == C)
&& (...

PR #133, opened 2020-03-19
72246 Get tags by tag functions proxy

janid1967:
We've earlier discussed that user should select pairs of TagFunctionCodres/RegisterCodes. We've today pinpointed that end user must work with (i.e select) one by one TagFunctions when searching for Main tags with that specific TagFunction. This because the wizard, at last page, should assist end user to assign requirement to Tags pr. TagFunction. This will not work if the page have input tags cross multiple TagFunctions

Also refactored query getting all tags in project to use IReadOnlyContext


janid1967:
Will probably be changed ..... we can't handle multiple TagFunctionCode/RegisterCode pairs each request. Closing and re-open later when fixed

src/Equinor.Procosys.Preservation.Query/TagApiQueries/SearchTags/SearchTagsByTagFunctionQueryHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+            var apiTags = await _tagApiService
+                .SearchTagsByTagFunction(_plantProvider.Plant, request.ProjectName, request.TagFunctionCode, request.RegisterCode)
+                ?? new List<ProcosysTagOverview>();
+
+            var presTagNos = await (from tag in _context.QuerySet<Tag>()
+                join p in _context.QuerySet<Project>() on EF.Property<int>(tag, "ProjectId") equals p.Id
+                where p.Name == request.ProjectName
+                select tag.TagNo).ToListAsync(cancellationToken);
+
+            // Join all tags from API with preservation tags on TagNo. If a tag is not in preservation scope, use default value (null).
```
TordJoranger: It has to be possible to make this a bit more readable? It took me 5 minutes just to figure out whats going on. Better variable names for x and y would help a lot I think (especially for the selectMany() second argument

janid1967: Agree, but I nearly don't understand it myself. H wrote similar code in other QueryHandler so I need assistance on that

janid1967: @hbjorgo Can You assist ? we have same code in handler when search tags ny tagno

PR #143, opened 2020-03-27
72144 endpoint for unique filter values area tag function dicipline

llykkelarsen:
Endpoint for unique filter values:
- Area codes
- Tag function codes
- Discipline codes

Tests

### DIFF
--- a/src/Equinor.Procosys.Preservation.Query/GetUniqueTagFunctions/TagFunctionCodeDto.cs
+++ b/src/Equinor.Procosys.Preservation.Query/GetUniqueTagFunctions/TagFunctionCodeDto.cs
@@ -2,10 +2,7 @@
 {
     public class TagFunctionCodeDto
     {
-        public TagFunctionCodeDto(string code)
-        {
-            Code = code;
-        }
+        public TagFunctionCodeDto(string code) => Code = code;
 
         public string Code { get; }
     }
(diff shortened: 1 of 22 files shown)


src/Equinor.Procosys.Preservation.Query/GetUniqueTagFunctions/TagFunctionCodeDto.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Equinor.Procosys.Preservation.Query.GetUniqueTagFunctions
+{
+    public class TagFunctionCodeDto
+    {
+        public TagFunctionCodeDto(string code)
+        {
+            Code = code;
```
janid1967: expr.body

src/tests/Equinor.Procosys.Preservation.Query.Tests/GetUniqueTagAreas/GetUniqueTagAreasQueryHandlerTests.cs
```diff
@@ -0,0 +1,73 @@
...
+                _currentUserProvider))
+            {
+                _testDataSet = AddTestDataSet(context);
+
+                _queryForProject1 = new GetUniqueTagAreasQuery(_testDataSet.Project1.Name);
+            }
+        }
+
+        [TestMethod]
+        public async Task HandleGetAllTagsInProjectQuery_ShouldReturnOkResult()
```
janid1967: The first part of method name do not reflect correct Query. Should be HandleGetUniqueTagAreasQuery_xxx on all methods in this class. I guess we have the same issue in all other "similar" test-classes. Fix all, I don't repeat this comment

janid1967:
merge it

PR #144, opened 2020-03-27
71434 Record values with explicit types

janid1967:
In today solution, client send double and bool values as strings when recording values for fields. 
This is a refactor of this, so client send actual double and bool values to avoid string parsing.

**Breaking change for client. This PR will not / can not be merged before Mars demo**


janid1967:
@kfaln This is the new code I talked about. Can You take a look at this PR and see if you can find time to take the UI part of it? UI PBI is 72371 (estimate 12h, but I guess You use less)

src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/PreservationPeriod.cs
```diff
@@ -106,36 +106,47 @@ public void SetComment(string comment)
...
+            if (field.FieldType != FieldType.Number)
             {
-                return;
+                throw new Exception($"Can't record a {nameof(FieldType.Number)} value for a {field.FieldType} field");
             }
+            ValidateAndPrepareForNewRecording(field);
 
-            switch (field.FieldType)
+            // save new value ONLY if there is a value!
+            if (value.HasValue)
```
kfaln, line 133: What happens when a user "blanks" out a number field? It seems this check prevents the new (blank) value from being stored. 

Any previous value is removed in `ValidateAndPrepareForNewRecording`.

Afterwards, when reading back the value in UI, I think the default `double` value 0 is returned.

janid1967: Have U tested it from client? The backend logic is: when we had a value, and blank comes in, the old value should be removed and no new record should be saved.
I think I have unit test for that case, but will double-check it

janid1967: I can't reproduce. I tested it against db in Azure on Tag with ID 52 (1301-E005). Using Swagger and recording for requirement with id 64 with this payload:

{
  "numberValues": [
    {
      "fieldId": 6,
      "value": 99,
      "isNA": false
    }
  ],
  "checkBoxValues": [],
  "comment": "none"
}

and afterwards

{
  "numberValues": [
    {
      "fieldId": 6,
      "value": null,
      "isNA": false
    }
  ],
  "checkBoxValues": [],
  "comment": "none"
}

... the 99-value are deleted from DB.

kfaln: I will do some more testing.

src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagsController.cs
```diff
@@ -273,13 +273,20 @@ public async Task<ActionResult<Query.CheckAreaTagNo.AreaTagDto>> CheckAreaTagNo(
             [FromRoute] int requirementId,
             [FromBody] RequirementValuesDto requirementValuesDto)
         {
-            var fieldValues = requirementValuesDto?
-                .FieldValues
-                .ToDictionary(
-                    keySelector => keySelector.FieldId,
-                    elementSelector => elementSelector.Value);
+            var numberValues = requirementValuesDto?
```
kfaln, line 276: The method name `RecordCheckBoxChecked` might not be entirely accurate...

src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/PreservationPeriod.cs
kfaln: My bad. Error was in UI.

PR #147, opened 2020-03-31
72335 change current plant requirement

llykkelarsen:
- remove default plant header
- manually add to all relevant endpoints
- create constants class
- edit tests to include new parameter. (this turned out to be a lot, let me know if you guys have a better idea on how to go about this)


janid1967:
do not merge. Henning will contact you

PR #149, opened 2020-04-01
72335 new change current plant requirement

llykkelarsen:
remove default plant header
manually add to all relevant endpoints
create constants class
adjust tests to fit changes


src/Equinor.Procosys.Preservation.Domain/Constants.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Equinor.Procosys.Preservation.Domain
```
janid1967, line 1: delete this class
Use PlantEntityBase.PlantLengthMax instead of your MaxLenth
Extend PlantEntityBase with PlantLengthMin = 5

src/Equinor.Procosys.Preservation.WebApi/Controllers/Responsibles/ResponsiblesController.cs
```diff
@@ -1,6 +1,9 @@
 ﻿using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Threading.Tasks;
+using Equinor.Procosys.Preservation.Domain;
 using Equinor.Procosys.Preservation.Query.ResponsibleAggregate;
+using Equinor.Procosys.Preservation.WebApi.Misc;
```
janid1967, line 6: Change namespace in this class. Don't match foldername (Responsibles)

src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagSearchController.cs
```diff
@@ -25,7 +28,13 @@ public class TagSearchController : ControllerBase
         /// <param name="startsWithTagNo"></param>
         /// <returns>All ProCoSys tags that match the search parameters</returns>
         [HttpGet]
-        public async Task<ActionResult<List<ProcosysTagDto>>> SearchTagsByTagNo([FromQuery] string projectName, [FromQuery] string startsWithTagNo)
+        public async Task<ActionResult<List<ProcosysTagDto>>> SearchTagsByTagNo(
+            [FromHeader( Name = PlantProvider.PlantHeader)]
+            [Required]
+            [StringLength(Constants.Plant.MaxLength, MinimumLength = Constants.Plant.MinLength)]
+            string plant,
```
janid1967, line 35: Add xml-doc comment (seems that this class is the only one with such comments, and my Resharper complains)

src/Equinor.Procosys.Preservation.Domain/Constants.cs
llykkelarsen: the max length in PlantEntityBase is 255 and in library it's 1024, should they be equal?

janid1967: Actually it should be 30! This because our Plant-column is the "same" column as PROJECTSCHEMA column on all tables in ProCosys Oracle DB. So any value wont ever be longer than length there (30).
But if you change to 30, you must also create a migration since it will change the DB.

Therefore we keep 255 for now (maybe a TECH to change to 30 afterwards since it should be done in Hennings Library code too)

PR #150, opened 2020-04-01
71187 Add procosys permissions as role claims

janid1967:
Added ClaimsTransformation, which gets procosys permissions from IPermissionService.cs (either from cache or calling Main WebAPI)
Used AuthorizeAttribute on JourneysController only to proof it works.

Created an IOperationFilter AddRoleDocumentation which generates documentation in swagger:
Created doc is for each endpoint with Autorize attr.
Doc say:
**401: User is not authenticated
403: User does not have the required permissions (PermA | PermB) & (PermC)**

Need to talk to Cato to define which permission (role) to secure which endpoint.

### DIFF
--- a/src/Equinor.Procosys.Preservation.WebApi/Middleware/ApiDescriptionExtension.cs
+++ b/src/Equinor.Procosys.Preservation.WebApi/Middleware/ApiDescriptionExtension.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
+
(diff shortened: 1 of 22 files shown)


src/Equinor.Procosys.Preservation.WebApi/Middleware/ApiDescriptionExtension.cs
```diff
@@ -0,0 +1,22 @@
...
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
+
+namespace Equinor.Procosys.Preservation.WebApi.Middleware
+{
+    public static class ApiDescriptionExtension
+    {
+        public static AuthorizeAttribute[] GetAuthorizeAttributes(this ApiDescription apiDescription)
+        {
+            var hasAuthAttribute = apiDescription.ActionDescriptor.EndpointMetadata
+                //.Select(filterInfo => filterInfo.Instance)
```
Cnordbo, line 15: Remove comment

PR #151, opened 2020-04-02
72393 project access

janid1967:
POC - want comments on this before working more. Not to be merged yet, but want POC approval


src/Equinor.Procosys.Preservation.WebApi/ProjectAccess/ProjectHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public class ProjectHelper : IProjectHelper
+    {
+        private readonly IReadOnlyContext _context;
+
+        public ProjectHelper(IReadOnlyContext context) => _context = context;
+
+        public async Task<string> GetProjectNameFromTagIdAsync(int tagId)
+        {
+            var projectName = await (from p in _context.QuerySet<Project>()
```
paal123, line 17: It might actually be feasible to add the projectname directly to tag-object to improve the performance on this query to avoid the join. Projectname is never changed anyway? right?

src/Equinor.Procosys.Preservation.WebApi/ProjectAccess/ProjectAccessValidator.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public async Task<bool> ValidateAsync<TRequest>(TRequest request) where TRequest : IBaseRequest
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            // If ProjectAccessCheckAttribute found on request, user need access to Project to allow request execution
+            var projectAccessCheck = request.GetType().GetCustomAttribute<ProjectAccessCheckAttribute>();
```
paal123, line 28: Alternatively, we could create a new interface which implements IBaseRequest (or a base class), and let our new interace keep track of these thing, to avoid the reflection part. But I dont know if it makes it much better...

src/Equinor.Procosys.Preservation.WebApi/ProjectAccess/ProjectHelper.cs
janid1967: That's correct, but we've chosen that Project is the Aggregate root, and Tag is a related object under Project.. Changing this now will be a massive change

src/Equinor.Procosys.Preservation.WebApi/ProjectAccess/ProjectAccessValidator.cs
janid1967: Can we talk over sample code what you mean?

janid1967: See commit

src/Equinor.Procosys.Preservation.WebApi/ProjectAccess/ProjectHelper.cs
janid1967: Another thing: The Project table contains other info than ProjectName (IsClosed for instance)

paal123:
So far so good!

PR #159, opened 2020-04-06
72553 - Added permissions to each endpoint

janid1967:


src/Equinor.Procosys.Preservation.WebApi/Controllers/TagFunctions/TagFunctionsController.cs
```diff
@@ -46,7 +49,7 @@ public class TagFunctionsController : ControllerBase
             return this.FromResult(result);
         }
 
-        [HttpPut]
```
llykkelarsen, line 49: Is this intentional?

janid1967: Is  [HttpPut] removed?

janid1967: Fixed

PR #167, opened 2020-04-15
72390 tags auto scope

janid1967:
Logic: All tags POST'ed to this new endpoint must have a TagFunction/Register-combination where preservation DB already has a TagFunction with that TagFunction/Register-combination. That TagFunction MUST have at least one non-voided Requirement.

The handler uses these TagFunction requirements as "template" and create a Tag requirement foreach of them

(Prior client uses this POST endpoint, the client already have used these TagFunctions with preservation requirements to search ProCoSys for tags with valid TagFunction/Register-combination to be able to autoscope tags)


src/Equinor.Procosys.Preservation.Command/TagCommands/AutoScopeTags/AutoScopeTagsCommandHandler.cs
```diff
@@ -0,0 +1,177 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Equinor.Procosys.Preservation.Domain;
+using Equinor.Procosys.Preservation.Domain.AggregateModels.JourneyAggregate;
+using Equinor.Procosys.Preservation.Domain.AggregateModels.ProjectAggregate;
+using Equinor.Procosys.Preservation.Domain.AggregateModels.RequirementTypeAggregate;
+using Equinor.Procosys.Preservation.Domain.AggregateModels.TagFunctionAggregate;
+using TagRequirement = Equinor.Procosys.Preservation.Domain.AggregateModels.ProjectAggregate.Requirement;
```
paal123, line 10: There is a design flaw in our solutions when we have to use these kind of using statements. This should be solve the correct way, and not use these kinds of using statements, as my experiences are that they tend to cause problems in the future for us.

src/Equinor.Procosys.Preservation.Command/TagCommands/AutoScopeTags/AutoScopeTagsCommandHandler.cs
```diff
@@ -0,0 +1,177 @@
...
+            _requirementTypeRepository = requirementTypeRepository;
+            _unitOfWork = unitOfWork;
+            _plantProvider = plantProvider;
+            _tagApiService = tagApiService;
+        }
+
+        public async Task<Result<List<int>>> Handle(AutoScopeTagsCommand request, CancellationToken cancellationToken)
+        {
+            var step = await _journeyRepository.GetStepByStepIdAsync(request.StepId);
+            var tagDetailList = await _tagApiService.GetTagDetailsAsync(_plantProvider.Plant, request.ProjectName, request.TagNos);
```
paal123, line 48: do not await getting steps and tagDetailsList, rather wait before executing the GetNeededTagFunctionsWithRequirementsAsync

src/Equinor.Procosys.Preservation.Command/TagCommands/AutoScopeTags/AutoScopeTagsCommandHandler.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            var tagFunctionsWithRequirements = await GetNeededTagFunctionsWithRequirementsAsync(tagDetailList);
+
+            var reqDefIds = tagFunctionsWithRequirements
+                .SelectMany(r => r.Requirements)
+                .Where(r => !r.IsVoided)
+                .Select(r => r.RequirementDefinitionId)
+                .Distinct()
+                .ToList();
+            var reqDefs = await _requirementTypeRepository.GetRequirementDefinitionsByIdsAsync(reqDefIds);
```
paal123, line 58: Same with this one

src/Equinor.Procosys.Preservation.Command/TagCommands/AutoScopeTags/AutoScopeTagsCommandHandler.cs
```diff
@@ -0,0 +1,177 @@
...
+                TagFunctionCode = tagDetails.TagFunctionCode
+            };
+
+            tag.SetArea(tagDetails.AreaCode, tagDetails.AreaDescription);
+            tag.SetDiscipline(tagDetails.DisciplineCode, tagDetails.DisciplineDescription);
+            
+            return tag;
+        }
+
+        class ProcosysTagDetailsComparer : IEqualityComparer<ProcosysTagDetails>
```
paal123, line 148: Might consider moving this class to another place and avoid as inner class

src/Equinor.Procosys.Preservation.Command/TagCommands/CreateTags/CreateTagsCommandHandler.cs
```diff
@@ -71,9 +73,10 @@ public async Task<Result<List<int>>> Handle(CreateTagsCommand request, Cancellat
             return new SuccessResult<List<int>>(addedTags.Select(t => t.Id).ToList());
         }
 
-        private async Task<Tag> CreateTagAsync(
-            ProcosysTagDetails tagDetails,
+        private Tag CreateTag(
```
paal123, line 76: Will not the requirements list be rebuild for each tag even though it is always the same?

src/Equinor.Procosys.Preservation.Command/TagCommands/AutoScopeTags/AutoScopeTagsCommandHandler.cs
janid1967: EF Core do not support multiple threads in same context. We've tried and it fails with message saying exactly that

src/Equinor.Procosys.Preservation.Command/TagCommands/CreateTags/CreateTagsCommandHandler.cs
janid1967: Yes, the list will be made for each Tag and it is supposed to be so. A row in the Requirement table, is a many-to-many connection between a Tag and a Requirement definition

src/Equinor.Procosys.Preservation.Command/TagCommands/AutoScopeTags/AutoScopeTagsCommandHandler.cs
janid1967: Agree, but it is a massive job. A rename of the Requirement table to TagRequirement. will also leads to a rename of the DB table (plus migration).

I rather make a TECH on that and do it in sep PR. OK @paal123 ?

janid1967: renanamed

src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagsController.cs
```diff
@@ -22,7 +23,7 @@
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using ServiceResult.ApiExtensions;
-using Requirement = Equinor.Procosys.Preservation.Command.Requirement;
+using RequirementForCommand = Equinor.Procosys.Preservation.Command.RequirementForCommand;
```
paal123, line 26: could just be using Equinor.Procosys.Preservation.Command namespace but never mind

janid1967: I fix

PR #168, opened 2020-04-15
72878 put endpoint for tag

llykkelarsen:
endpoint for updating tag
validators
tests


src/tests/Equinor.Procosys.Preservation.Command.Tests/TagCommands/UpdateTag/UpdateTagCommandTests.cs
```diff
@@ -0,0 +1,19 @@
...
+    [TestClass]
+    public class UpdateTagCommandTests
+    {
+        [TestMethod]
+        public void Constructor_ShouldSetProperties()
+        {
+            var dut = new UpdateTagCommand(2, "Remark", "StorageArea");
+
+            Assert.AreEqual(2, dut.TagId);
+            Assert.AreEqual("Remark", dut.Remark);
```
paal123, line 15: Use read only string similar to other tests

janid1967: Fine for me @paal123, but in such simple constructor test where the Assert is right below (and no other ref to same string), I find this solution fine. It's like this in every contructor test in solution ... I guess it's hundreds of them

llykkelarsen: won't fix

PR #173, opened 2020-04-20
71254 optimistic concurrency

llykkelarsen:
Setting the solution up for optimistic concurrency in order to handle conflicts faced with two or more requests updating the same object or the same entity instance.

As of right now this code does not interfere with any of the existing functionality.

### DIFF
--- a/src/Equinor.Procosys.Preservation.Domain/EntityBase.cs
+++ b/src/Equinor.Procosys.Preservation.Domain/EntityBase.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using MediatR;
 
 namespace Equinor.Procosys.Preservation.Domain
@@ -11,7 +12,21 @@ namespace Equinor.Procosys.Preservation.Domain
         private List<INotification> _domainEvents;
 
         public IReadOnlyCollection<INotification> DomainEvents => _domainEvents?.AsReadOnly() ?? (_domainEvents = new List<INotification>()).AsReadOnly();
+
         public virtual int Id { get; protected set; }
+
+        public byte[] RowVersion { get; protected set; }
+
+        public ulong GetRowVersion() => (ulong) BitConverter.ToInt64(RowVersion);
+        public void SetRowVersion(ulong version)
+        {
+            var newRowVersion = BitConverter.GetBytes(version);
+
+            for (var index = 0; index < newRowVersion.Length; index++)
+            {
+                RowVersion[index] = newRowVersion[index];
+            }
+        }
 
         public void AddDomainEvent(INotification eventItem)
         {
--- a/src/Equinor.Procosys.Preservation.Infrastructure/Migrations/20200420085753_AddedRowVersion.cs
+++ b/src/Equinor.Procosys.Preservation.Infrastructure/Migrations/20200420085753_AddedRowVersion.cs
@@ -1,0 +1,184 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Equinor.Procosys.Preservation.Infrastructure.Migrations
+{
+    public partial class AddedRowVersion : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<byte[]>(
+                name: "RowVersion",
+                table: "Tags",
+                rowVersion: true,
+                nullable: true);
+
+            migrationBuilder.AddColumn<byte[]>(
+                name: "RowVersion",
+                table: "TagRequirements",
+                rowVersion: true,
+                nullable: true);
+
+            migrationBuilder.AddColumn<byte[]>(
+                name: "RowVersion",
+                table: "TagFunctions",
+                rowVersion: true,
+                nullable: true);
+
+            migrationBuilder.AddColumn<byte[]>(
+                name: "RowVersion",
+                table: "TagFunctionRequirements",
+                rowVersion: true,
+                nullable: true);
+
+            migrationBuilder.AddColumn<byte[]>(
+                name: "RowVersion",
+                table: "Steps",
+                rowVersion: true,
+                nullable: true);
+
+            migrationBuilder.AddColumn<byte[]>(
+                name: "RowVersion",
+                table: "Responsibles",
+                rowVersion: true,
+                nullable: true);
+
+            migrationBuilder.AddColumn<byte[]>(
+                name: "RowVersion",
+                table: "RequirementTypes",
+                rowVersion: true,
+                nullable: true);
+
+            migrationBuilder.AddColumn<byte[]>(
+                name: "RowVersion",
+                table: "RequirementDefinitions",
+                rowVersion: true,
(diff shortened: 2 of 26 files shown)


paal123:
This will work fine as all saves will use row version currently on entity in db. So this code can be deployed as a step 1 without any problems!

src/Equinor.Procosys.Preservation.Domain/EntityBase.cs
```diff
@@ -11,8 +12,22 @@ public abstract class EntityBase
         private List<INotification> _domainEvents;
 
         public IReadOnlyCollection<INotification> DomainEvents => _domainEvents?.AsReadOnly() ?? (_domainEvents = new List<INotification>()).AsReadOnly();
+
         public virtual int Id { get; protected set; }
 
+        public byte[] RowVersion { get; protected set; }
+
+        public ulong GetRowVersion() => (ulong) BitConverter.ToInt64(RowVersion);
```
janid1967, line 20: Add a test for Get- SetRowVersion?

src/Equinor.Procosys.Preservation.Infrastructure/Migrations/20200420085753_AddedRowVersion.cs
```diff
@@ -0,0 +1,184 @@
...
+{
+    public partial class AddedRowVersion : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<byte[]>(
+                name: "RowVersion",
+                table: "Tags",
+                rowVersion: true,
+                nullable: true);
```
janid1967, line 14: So it had to be nullable anyway? Why?

llykkelarsen: It did, turns out the rowversion datatype value is never null by default.

src/tests/Equinor.Procosys.Preservation.Domain.Tests/EntityBaseTests.cs
```diff
@@ -50,9 +51,26 @@ public void DomainEventsAreClearedTest()
             Assert.AreEqual(0, dut.DomainEvents.Count);
         }
 
-        public class TestableEntityBase : EntityBase
+        [TestMethod]
+        public void GetRowVersion()
```
janid1967, line 55: _ShouldXXX

src/tests/Equinor.Procosys.Preservation.Domain.Tests/EntityBaseTests.cs
```diff
@@ -50,9 +51,26 @@ public void DomainEventsAreClearedTest()
...
+        {
+            var dut = new TestableEntityBase();
+            dut.SetProtectedRowVersion(123);
+            dut.GetRowVersion().Equals(123);
+        }
+
+        [TestMethod]
+        public void SetRowVersion()
```
janid1967, line 63: _ShouldXXX

src/tests/Equinor.Procosys.Preservation.Domain.Tests/EntityBaseTests.cs
```diff
@@ -50,9 +51,26 @@ public void DomainEventsAreClearedTest()
...
+        {
+            var dut = new TestableEntityBase();
+            dut.SetProtectedRowVersion(123);
+            dut.SetRowVersion(123);
```
janid1967, line 67: Maybe other than 123 both in both SetProtectedRowVersion and SetRowVersion

Should we demand increase version (and fail if setting lower)?

llykkelarsen: First part done. Won't fix for the second part because it is not our responsibility.

llykkelarsen:
The biggest issue I had with getting this to work was that it would fail if the entire byte array was replaced in the setting of row version. Hence the for loop to instead replace the content. I tried to write some tests exemplifying this but I am open to suggestions if there is a better (simpler) way? 
@janid1967  @hbjorgo @paal123

src/Equinor.Procosys.Preservation.Domain/EntityBase.cs
```diff
@@ -11,8 +12,22 @@ public abstract class EntityBase
         private List<INotification> _domainEvents;
 
         public IReadOnlyCollection<INotification> DomainEvents => _domainEvents?.AsReadOnly() ?? (_domainEvents = new List<INotification>()).AsReadOnly();
+
         public virtual int Id { get; protected set; }
 
+        public byte[] RowVersion { get; protected set; }
```
hbjorgo, line 18: Consider exposing the array as read only (https://docs.microsoft.com/en-us/dotnet/api/system.array.asreadonly?view=netstandard-2.1) to avoid others changing it.

src/tests/Equinor.Procosys.Preservation.Domain.Tests/EntityBaseTests.cs
```diff
@@ -50,9 +54,68 @@ public void DomainEventsAreClearedTest()
...
+            {
+                var dut = new TestableEntityBase();
+
+                dut.SetProtectedRowVersion(OldRowVersion);
+                var originalByteArray = dut.GetRowVersionPointer();
+
+                dut.SetProtectedRowVersion(NewRowVersion);
+                var newByteArray = dut.GetRowVersionPointer();
+
+                Assert.IsFalse(originalByteArray == newByteArray, "Reference equals should be false as we replace the byte array");
```
hbjorgo, line 104: Isn't this test testing the sub class (created for testing only) used?

src/Equinor.Procosys.Preservation.Domain/EntityBase.cs
hbjorgo: Consider restricting the setter to private to avoid sub classes messing with it directly instead of through the get/set methods.

llykkelarsen: I don't think RowVersion can be read only, we have to override it when we save?
If I make the setter private I am unable to write any tests on this. To test the set method there must be an existing RowVersion.

hbjorgo: I haven't tested my proposal, so it might not be possible.
To test it, you might need to call the set method twice in a row...

llykkelarsen: I'm pretty sure this is the only way to establish out of context concurrency with a web API. The first RowVersion will always come from Entity Framework which is why our SetRowVersion method won't work in my tests before it is set with a simplified version of the set method (SetProtectedRowversion). If you look at the SetRowVersion method, it won't matter how many times you call it, you simply can't set the property without an existing value.

paal123:
To save time, as we are on a tight schedule, I suggest we just leave it as it is. It is working, even though it does not use the readonly collection in the same way as other entity collections do, but in addition this is not a entity collection and is somewhat different that the rest.

If we get in trouble further down the road, we can consider switching to some kind of readonly collection or similar.

PR #176, opened 2020-04-21
73245 security on Content restrictions + Plant Cache

janid1967:
Logic:
- Get list of Content restrictions (i.e responsibleCodes) from main API. Cache them. (one and only one Content restriction with value = "%" means "No restrictions" in ProCoSys)
- Add UserData claim for each restrictions (inclusive %)
- When user perform a command on a Tag or item belonging to a item under Tag, it is validated if user has any restrictions, If so, user must have explicit access to content for responsiblecode found via current step of Tag
- ClaimsTransformation now exit early if either plant not given, user do not have plant access or oid claim not exist on principal.
- All api calls to main api for getting any type of data for building claims (project, permissions and restrictions), do not check in valid plant as before.
- Refactor all api calls to main api to use plant cache to check for valid plant instead of checking over wire each time


src/Equinor.Procosys.Preservation.WebApi/Caches/IPermissionCache.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Equinor.Procosys.Preservation.WebApi.Caches
+{
+    public interface IPermissionCache
```
paal123, line 7: Should we have an invalidate cache for current user? Would be good for us during testing at least, and it could be practical to have it the preservation app as well to be able to refresh to current rights, its 20 minutes lost work-time to wait for the cache to trigger a refresh

src/Equinor.Procosys.Preservation.WebApi/Misc/CurrentUserProvider.cs
```diff
@@ -12,12 +13,16 @@ public class CurrentUserProvider : ICurrentUserProvider
 
         public Guid GetCurrentUser()
         {
-            var oid = _accessor.HttpContext.User.Claims.TryGetOid();
-            if (oid.HasValue)
+            var guid = TryGetCurrentUserOid();
```
paal123, line 16: oid instead of guid

src/Equinor.Procosys.Preservation.WebApi/appsettings.json
```diff
@@ -30,7 +30,8 @@
   "TagOptions": {
     "TagIsNewHours": 48
   },
-  "PermissionOptions": {
-    "PermissionCacheMinutes": 20
+  "CacheOptions": {
```
paal123, line 33: Are these overwriting the values in the CacheOptions.cs in some way?

src/Equinor.Procosys.Preservation.WebApi/Authorizations/ContentRestrictionsChecker.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            var authenticatedUser = GetAuthenticatedUser();
+            if (authenticatedUser != null)
+            {
+                var userDataClaimWithContentRestriction = authenticatedUser.Claims.ContentRestrictionClaims();
+                return userDataClaimWithContentRestriction.Count > 1 && userDataClaimWithContentRestriction.First().Value !=
+                       ClaimsTransformation.NoRestrictions;
+            }
+
+            return false;
```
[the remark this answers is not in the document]
janid1967: May idea was normal flow: if you are not authenticated you don't have any restrictions.
If you're not authenticated you will be stopped anyway (so you will never come at this point)

src/Equinor.Procosys.Preservation.WebApi/Authorizations/AccessValidator.cs
```diff
@@ -0,0 +1,88 @@
...
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private async Task<bool> HasCurrentUserAccessToContentAsync(ITagCommandRequest tagCommandRequest)
+        {
+            if (_contentRestrictionsChecker.HasCurrentUserAnyRestrictions())
```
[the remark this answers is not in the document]
janid1967: I'll rewrite AccessValidator.cs as we discussed. Check if user is authenticated early in AccessValidator.cs. Return false if not. Methods in IProjectAccessChecker and IContentRestrictionsChecker can then throw exception if user is authenticated

src/Equinor.Procosys.Preservation.WebApi/Caches/IPermissionCache.cs
janid1967: How should we implement that? An endpoint?

paal123: Yes, a separate endpoint you can call to just, clear or rebuild your permissions. Takes no arguments (or maybe plant)

janid1967: Will be a TECH story

PR #177, opened 2020-04-22
first two endpoints for review

llykkelarsen:
rowversion changes for endpoints:
GET /Tags/{id}/Actions
GET /Tags/{id}/Actions/{actionId}

Would be great if someone could look over these and leave a comment so I know I'm on the right track for the rest.

### DIFF
--- a/src/Equinor.Procosys.Preservation.Query/GetTagActionDetails/GetActionDetailsQueryHandler.cs
+++ b/src/Equinor.Procosys.Preservation.Query/GetTagActionDetails/GetActionDetailsQueryHandler.cs
@@ -55,7 +55,8 @@ namespace Equinor.Procosys.Preservation.Query.GetTagActionDetails
                 dto.Action.DueTimeUtc,
                 dto.Action.IsClosed,
                 closedBy,
-                dto.Action.ClosedAtUtc);
+                dto.Action.ClosedAtUtc,
+                dto.Action.GetRowVersion());
             
             return new SuccessResult<ActionDetailsDto>(action);
         }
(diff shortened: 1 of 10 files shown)


src/Equinor.Procosys.Preservation.Query/GetTagActionDetails/GetActionDetailsQueryHandler.cs
```diff
@@ -55,7 +55,8 @@ public async Task<Result<ActionDetailsDto>> Handle(GetActionDetailsQuery request
                 dto.Action.DueTimeUtc,
                 dto.Action.IsClosed,
                 closedBy,
-                dto.Action.ClosedAtUtc);
+                dto.Action.ClosedAtUtc,
```
paal123, line 58: I know this is not something you have changed, but the word dto here is rather confusing as it is not used outside of this method, and is hence on a data transfer object.

janid1967: What do you suggest @paal123 .. just ActionDetails?

llykkelarsen:
I had a couple of failing tests due to the not being able to set rowversion issue (this will ultimately be the case for each new endpoint I change), so I made an extensionmethod for EntityBase allowing us to set rowversion in the test data from the get-go. @janid1967 @hbjorgo 
If you guys agree with this way of doing it, I will merge this and do continous, smaller PRs for the rest of the changes like Pål suggested above.

src/Equinor.Procosys.Preservation.Query/GetTagActionDetails/GetActionDetailsQueryHandler.cs
paal123: Using dtos are correct. It was just a minor comment on the variable dto here, since the variable never leaves the method, and is replaced by the ActionDetailsDto. 

It is just these lines I don't like:
select new Dto
                 {
                     Action = a,
                     CreatedBy = createdUser,
                     ClosedBy = closedUser
                 }

paal123: @janid1967 I would have used an anonymous class and called the variable result or something similar, just to make sure its easier to see that it is not an dot object that is sent out of the method

PR #178, opened 2020-04-23
BUGS on CheckAreaTagNo

janid1967:
Bug1: Endpoint didn't have plant as required param
Bug2: Query didn't have IProjectRequest interface (which cause security on Project)
Improvement: Validator -> swagger show which params are required

### DIFF
--- a/src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/AreaTagDto.cs
+++ b/src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/AreaTagDto.cs
@@ -3,7 +3,7 @@
     public class AreaTagDto
     {
         public string ProjectName { get; set; }
-        public AreaTagType AreaTagType { get; set; } = AreaTagType.PreArea;
+        public AreaTagType AreaTagType { get; set; }
         public string DisciplineCode{ get; set; }
         public string AreaCode{ get; set; }
         public string TagNoSuffix{ get; set; }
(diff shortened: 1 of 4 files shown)


src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/AreaTagDto.cs
```diff
@@ -3,7 +3,7 @@
     public class AreaTagDto
     {
         public string ProjectName { get; set; }
-        public AreaTagType AreaTagType{ get; set; }
+        public AreaTagType AreaTagType { get; set; } = AreaTagType.PreArea;
```
Cnordbo, line 6: default needed / correct?

Cnordbo: AreaTagType cant be null in the validator, so its required by the API.

janid1967: > default needed / correct?

It works without this. But when setting default value as this, it's pre-selected in swagger.

janid1967: > AreaTagType cant be null in the validator, so its required by the API.

I don't understand what u mean

Cnordbo: The point was that its required by the API input, and so a default value is not needed. 

One thing is to set it for swagger, but this could have the potential effect on any usage from external parties that its automatically set to "PreArea" when they forget to implement AreaTagType as a parameter?

This could potentially lead to unwanted sideeffect?

janid1967: Now I see. And agree, I'll remove default, push and merge

PR #179, opened 2020-04-24
72421 attachments table

janid1967:
First version of Attachment table. May be changed when integrating with blobstorage.
Main idea: Common Attachment table used for different entity types via Discriminator. For now, entities Tag and Action.

Next steps after this PR: 
- endpoint for adding Tag Attachment (without blob for now) 
- endpoint for listing all Tag Attachments 

(Do not use to much time to reflect over class ProjectAggregate/Attachment.cs which is renamed to AttachmentValue.cs here. AttachmentValue.cs is for recording av value of type attachment instead of type checkbox or number. Implementation of AttachmentValue.cs will later be extended/changed to also be integrated with a blob)


janid1967:
Will come a commit with unit tests soon

src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/ActionAttachment.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+
+namespace Equinor.Procosys.Preservation.Domain.AggregateModels.ProjectAggregate
```
hbjorgo, line 3: Are the sub classes necessary? What if there's just one class - Attachment?

src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/ActionAttachment.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public class ActionAttachment : Attachment
```
hbjorgo, line 5: Are the sub classes necessary? What if there's just one class - Attachment?

janid1967: You're probably right. See latest commits

janid1967: ....or not!!! When update database I now get Attachment table with both TagId and ActionId foregin key columns and both are non-nullable. This is the part we get "for free" whit abstract class Attachment, subclass for ActionAttachment and TagAttachment and and Discrimnator column

janid1967: I suggest I revert back 3 commits. OK @hbjorgo ?

PR #184, opened 2020-04-27
Blob service

hbjorgo:
Blob storage interface and implementation for Azure Blob Storage.
Tried to keep the interface generic, so that it's possible to use another storage medium (e.g. disk) if desirable.
SAS token generation is perhaps not applicable for all storage mediums, but still I kept all methods in the same interface.


janid1967:
I guess you need to fix something in build pipeline

PR #186, opened 2020-04-28
Created Update Mode endpoint

demissie:


src/tests/Equinor.Procosys.Preservation.Command.Tests/ModeCommands/UpdateMode/UpdateModeCommandHandlerTests.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            _dut = new UpdateModeCommandHandler(
+                modeRepositoryMock.Object,
+                UnitOfWorkMock.Object);
+        }
+
+        [TestMethod]
+        public async Task HandlingUpdateModeCommand_ShouldUpdateMode()
+        {
+            // Act
```
janid1967, line 41: This is fine, but can be done even better. See "same" test for UpdateStepCommandHandlerTest. There we Assert that title is _oldTitle before we change to _newTitle. This proof the code more.
So: Make an Assert at line 41 here to check for _oldTitle. Do the same in UpdateJourneyCommandHandlerTest (I guess you used that as "howto")

PR #187, opened 2020-04-28
73178 refactor tag validators for start and stop

janid1967:
When using StartPreservation command when onboarding Demissie for making StopPreservation, I found that validation logic for checking IsReadyToBeXXXX could be made better to reuse same logic in Tag domain class. Before this PR "same" logic was duplicated both in Tag.cs and in validators

### DIFF
--- a/src/Equinor.Procosys.Preservation.Command/Validators/TagValidators/TagValidator.cs
+++ b/src/Equinor.Procosys.Preservation.Command/Validators/TagValidators/TagValidator.cs
@@ -42,6 +42,7 @@ namespace Equinor.Procosys.Preservation.Command.Validators.TagValidators
             return tag != null && tag.Requirements.Any(r => !r.IsVoided);
         }
 
+        // todo remove after codereview .. diff get crazy when removing it
         public async Task<bool> AllRequirementDefinitionsExistAsync(int tagId, CancellationToken cancellationToken)
         {
             var tag = await GetTagWithRequirements(tagId, cancellationToken);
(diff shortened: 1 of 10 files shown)


src/Equinor.Procosys.Preservation.Command/Validators/TagValidators/TagValidator.cs
```diff
@@ -42,6 +42,7 @@ public async Task<bool> HasANonVoidedRequirementAsync(int tagId, CancellationTok
             return tag != null && tag.Requirements.Any(r => !r.IsVoided);
         }
 
+        // todo remove after codereview .. diff get crazy when removing it
```
llykkelarsen, line 45: don't forget to remove

src/Equinor.Procosys.Preservation.Command/Validators/TagValidators/TagValidator.cs
```diff
@@ -83,6 +84,7 @@ public async Task<bool> HasRequirementWithActivePeriodAsync(int tagId, int requi
             return requirement != null && requirement.HasActivePeriod;
         }
 
+        // todo remove after codereview .. diff get crazy when removing it
```
llykkelarsen, line 87: and here

src/Equinor.Procosys.Preservation.Command/Validators/TagValidators/TagValidator.cs
```diff
@@ -113,6 +115,7 @@ public async Task<bool> RequirementIsReadyToBePreservedAsync(int tagId, int requ
             return requirement != null && requirement.ReadyToBePreserved;
         }
 
+        // todo remove after codereview .. diff get crazy when removing it
```
llykkelarsen, line 118: and here

src/Equinor.Procosys.Preservation.Command/Validators/TagValidators/TagValidator.cs
janid1967: done

PR #188, opened 2020-04-28
72421 upload anf list tag attachment end points

janid1967:
First endpoints for uploading attachment to Tag and get list of Tag attachments
(No integration with Blob storage yet > no streams involved)


src/Equinor.Procosys.Preservation.Domain/AggregateModels/ProjectAggregate/Tag.cs
```diff
@@ -287,6 +287,10 @@ public void SetModified(Person modifiedBy)
         public bool IsReadyToBeStarted()
             => Status == PreservationStatus.NotStarted && Requirements.Any(r => !r.IsVoided);
 
+        public TagAttachment GetAttachmentByFileName(string fileName) =>
+            _attachments.SingleOrDefault(a =>
+                string.Compare(a.FileName, fileName, StringComparison.InvariantCultureIgnoreCase) == 0);
```
hbjorgo, line 292: Is there a special need to use invariant culture here, compared to other places?

janid1967: Changed to CurrentCultureIgnoreCase. Agree?
We must have IgnoreCase

janid1967: Changed to a.FileName.ToUpper() == fileName.ToUpper() as discussed

PR #191, opened 2020-04-29
73348 unvoid tag endpoint

demissie:


src/Equinor.Procosys.Preservation.Command/TagCommands/UnvoidTag/UnvoidTagCommand.cs
```diff
@@ -0,0 +1,15 @@
+﻿using MediatR;
+using ServiceResult;
+
+namespace Equinor.Procosys.Preservation.Command.TagCommands.UnvoidTag
+{
+    public class UnvoidTagCommand : IRequest<Result<Unit>>, ITagCommandRequest
+    {
+        public UnvoidTagCommand(
```
janid1967, line 8: make one-liner with body expression. I commented the same in Void tag earlier today. Do it there also if not done yet:

`public UnvoidTagCommand(int tagId) => TagId = tagId;`

src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagsController.cs
```diff
@@ -451,6 +452,20 @@ public class TagsController : ControllerBase
             return this.FromResult(result);
         }
 
+        [Authorize(Roles = Permissions.PRESERVATION_WRITE)]
```
janid1967, line 455: Change permission to Permissions.PRESERVATION_PLAN_WRITE. Same for Void

PR #199, opened 2020-05-05
Dockerfile

hbjorgo:
Running the app as a specified user in container. Exposing port 5000.


janid1967:
It has little value that I review such. Christer 1 or Christer 2 should

PR #222, opened 2020-05-14
Uniform more row version tests

janid1967:
Yesterday we found a better way to test RowVersion flow in Handlers.
This PR is to uniform these tests even more.

**PS: Writing these tests, I also found bug in both CloseActionCommandHandler.cs and UpdateActionCommandHandler.cs. They returned RowVersion.ToString(), which is "System.Byte[]"**


janid1967:
@demissie After this PR is merged into Master, you should merge master into your working branch for Void/Unvoid Journey and update Handler-tests in same way as in this PR

llykkelarsen:
I've updated the tests for update requirements, transfer, and complete preservation. Only ones missing now are void/unvoid, @demissie since these are yours if you could fix VoidTagCommandHandlerTests.cs & UnvoidTagCommandHandlerTests.cs that would be great.

janid1967:
Looking ok @Can you approve @demissie and @llykkelarsen ... I can't since I've created it

janid1967:
I merge now. Demissie can fix in other brach

PR #223, opened 2020-05-14
73443 delete attachment from tag

janid1967:
Also found that ActionValidator missed unit tests, which again lead to a bug in ActionValidator: No need to merge Tag to check for exists on Action by its id. Fixed


src/tests/Equinor.Procosys.Preservation.Command.Tests/TagAttachmentCommands/Delete/DeleteTagAttachmentCommandValidatorTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        private Mock<ITagValidator> _tagValidatorMock;
+        private Mock<IAttachmentValidator> _attachmentValidatorMock;
+        private DeleteTagAttachmentCommand _command;
+
+        [TestInitialize]
+        public void Setup_OkState()
+        {
+            _projectValidatorMock = new Mock<IProjectValidator>();
+
+            _command = new DeleteTagAttachmentCommand(2, 3, "AAAAAAAAABA=");
```
llykkelarsen, line 25: could use null for rowversion, the value is not used in these tests

PR #235, opened 2020-05-19
73448 API requirement definition endpoints void / unvoid

demissie:


src/Equinor.Procosys.Preservation.WebApi/Controllers/RequirementTypes/UnvoidRequirementDefinitionDto.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Equinor.Procosys.Preservation.WebApi.Controllers.RequirementType
```
janid1967, line 1: Plural ... since you've renamed folder (i.e namespace), you must rename namespace in all classes inside ns. I see 3 of them

PR #240, opened 2020-05-22
73923 attachment count for get action dto

llykkelarsen:


src/Equinor.Procosys.Preservation.Query/GetActionDetails/ActionDetailsDto.cs
```diff
@@ -14,7 +14,8 @@ public class ActionDetailsDto
             bool isClosed,
             PersonDto closedBy,
             DateTime? closedAtUtc,
-            string rowVersion)
+            string rowVersion,
```
janid1967, line 17: rowVersion as last param since this is the "standard" other in dtos?

src/tests/Equinor.Procosys.Preservation.Query.Tests/GetActionDetails/GetActionDetailsQueryHandlerTests.cs
```diff
@@ -31,10 +32,14 @@ protected override void SetupNewDatabase(DbContextOptions<PreservationContext> d
                 _testDataSet = AddTestDataSet(context);
 
                 var tag = _testDataSet.Project1.Tags.First();
+                var attachment = new ActionAttachment(TestPlant, Guid.NewGuid(), "FileA");
+                var attachment2 = new ActionAttachment(TestPlant, Guid.NewGuid(), "FileB");
```
janid1967, line 36: remove attachment2. Then you will test for both zero and 1 attachment

src/Equinor.Procosys.Preservation.Query/GetActionDetails/GetActionDetailsQueryHandler.cs
```diff
@@ -68,6 +69,7 @@ private class Dto
             public Action Action { get; set; }
             public Person CreatedBy { get; set; }
             public Person ClosedBy { get; set; }
+            public int AttachmentCount { get; set; }
```
paal123, line 72: @llykkelarsen @janid1967 I would suggest removing the whole Dto class here. It is only a private class that lives inside one method. Suggest we rather use an anonymous type. By removing the whole Dto class and 
Replacing "select new Dto" with "select new", everything else can remain the same.

llykkelarsen: Fine with me. @janid1967 Thoughts?

janid1967: Ok. Demissie did the same in other PR

llykkelarsen:
My last commit was missing some changes, don't bother to comment till my next one.

PR #241, opened 2020-05-22
73746 API expanded action details endpoint

demissie:


src/tests/Equinor.Procosys.Preservation.Query.Tests/GetActionDetails/GetActionDetailsQueryHandlerTests.cs
```diff
@@ -131,5 +132,16 @@ private static void AssertAction(ActionDetailsDto actionDetailsDto, Action actio
             Assert.AreEqual(action.ClosedAtUtc, actionDetailsDto.ClosedAtUtc);
             Assert.AreEqual(isClosed, actionDetailsDto.IsClosed);
         }
+
+        private void AssertModifiedAction(Action action, Person modifier)
+        {
+            action.SetModified(modifier);
```
janid1967, line 138: you can't do this. Code line 138 is an Arrange to be able to test Asserts below
What I guess you can do is to perform 2 saves in Setup. The first save will set Created data. Second should set Modified data. If you can not assert exact modified time, asserting NotNull on modified time is alright

src/tests/Equinor.Procosys.Preservation.Query.Tests/GetActionDetails/GetActionDetailsQueryHandlerTests.cs
```diff
@@ -135,13 +139,10 @@ private static void AssertAction(ActionDetailsDto actionDetailsDto, Action actio
 
         private void AssertModifiedAction(Action action, Person modifier)
         {
-            action.SetModified(modifier);
-
             Assert.IsNotNull(action.ModifiedById);
             Assert.AreEqual(modifier.Id, action.ModifiedById);
             Assert.AreEqual(action.ModifiedAtUtc, action.ModifiedAtUtc);
```
janid1967: What do You really test here!? I guess 1 always equals 1 also

src/tests/Equinor.Procosys.Preservation.Query.Tests/GetActionDetails/GetActionDetailsQueryHandlerTests.cs
```diff
@@ -61,6 +61,7 @@ public async Task Handler_ReturnsClosedAction()
                 var actionDetailDto = result.Data;
 
                 AssertClosedAction(actionDetailDto, _closedAction, _testDataSet.CurrentUser);
+                AssertModifiedAction(_openAction, _testDataSet.CurrentUser);
```
janid1967: Here you Assert on object used for setup, not on object returned from Handler!

janid1967:
See my latest commit @demissie ... there you have proper test on modifiedby/time

src/Equinor.Procosys.Preservation.Query/GetActionDetails/GetActionDetailsQueryHandler.cs
```diff
@@ -68,6 +79,7 @@ private class Dto
             public Action Action { get; set; }
             public Person CreatedBy { get; set; }
             public Person ClosedBy { get; set; }
+            public Person ModifiedBy { get; set; }
```
paal123, line 82: Remove the whole "private class Dto" and replace by an anonymous type.

PR #242, opened 2020-05-25
Renaming routes

paal123:
Renaming endoints to plural
https://statoildeveloper.visualstudio.com/Statoil.NextGenProCoSys/_workitems/edit/73879

### DIFF
--- a/src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagsController.cs
+++ b/src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagsController.cs
@@ -488,7 +488,7 @@ namespace Equinor.Procosys.Preservation.WebApi.Controllers.Tags
         }
 
         [Authorize(Roles = Permissions.PRESERVATION_WRITE)]
-        [HttpPost("{id}/Requirement/{requirementId}/RecordValues")]
+        [HttpPost("{id}/Requirements/{requirementId}/RecordValues")]
         public async Task<IActionResult> RecordValues(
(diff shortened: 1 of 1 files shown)


src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagsController.cs
```diff
@@ -488,7 +488,7 @@ public class TagsController : ControllerBase
         }
 
         [Authorize(Roles = Permissions.PRESERVATION_WRITE)]
-        [HttpPost("{id}/Requirement/{requirementId}/RecordValues")]
+        [HttpPost("{id}/Requirements/{requirementId}/RecordValues")]
```
janid1967, line 491: This Route is in use in client. Should be changed at same time as merging this PR into master

src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagsController.cs
```diff
@@ -562,7 +562,7 @@ public class TagsController : ControllerBase
         }
 
         [Authorize(Roles = Permissions.PRESERVATION_WRITE)]
-        [HttpPost("{id}/Requirement/{requirementId}/Preserve")]
+        [HttpPost("{id}/Requirements/{requirementId}/Preserve")]
```
janid1967, line 565: This Route is in use in client. Should be changed in https://github.com/equinor/procosys-js-frontend at same time as merging this PR into master

PR #253, opened 2020-05-29
Missing property in tag function dto

janid1967:


src/Equinor.Procosys.Preservation.Query/GetTagFunctionDetails/RequirementDto.cs
```diff
@@ -2,13 +2,15 @@
 {
     public class RequirementDto
     {
-        public RequirementDto(int id, int requirementDefinitionId)
+        public RequirementDto(int id, int requirementDefinitionId, int interValWeeks)
         {
             Id = id;
             RequirementDefinitionId = requirementDefinitionId;
+            InterValWeeks = interValWeeks;
```
Cnordbo, line 9: Interval is a single word. no need to camelcase.

PR #272, opened 2020-06-09
add history table

llykkelarsen:


janid1967:
Ready for review or are you working on ObjectType / ObjectId?

llykkelarsen:
Ready for review.

src/Equinor.Procosys.Preservation.Domain/AggregateModels/HistoryAggregate/History.cs
```diff
@@ -0,0 +1,46 @@
...
+    public class History : PlantEntityBase, IAggregateRoot, ICreationAuditable
+    {
+        public const int DescriptionLengthMax = 1024;
+
+        protected History()
+            : base(null)
+        {
+        }
+
+        public History(
```
janid1967, line 17: both ObjectId + ObejctType need to come via constructor

src/Equinor.Procosys.Preservation.Domain/AggregateModels/HistoryAggregate/EventType.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Equinor.Procosys.Preservation.Domain.AggregateModels.HistoryAggregate
+{
+    public enum EventType
+    {
+        AddRequirement,
```
janid1967, line 5: Add PreserveRequirement

src/Equinor.Procosys.Preservation.Domain/AggregateModels/HistoryAggregate/History.cs
```diff
@@ -0,0 +1,50 @@
...
+            EventType = eventType;
+        }
+
+        public string Description { get; private set; }
+        public int CreatedById { get; private set; }
+        public int ObjectId { get; private set; }
+        public DateTime CreatedAtUtc { get; private set; }
+        public EventType EventType { get; private set; }
+        public ObjectType ObjectType { get; private set; }
+
```
janid1967, line 37: Add int PreservationRecordId. Use private setter. 
Create another constructor which take about same params as existing constructor. In new constructor, replace EventType param with PreservationRecord as param. Call constructor 1 from 2 with EventType.PreserveRequirement for EventType. Constructor 2 just keep PreservationRecord.id (thow exception if plant is uneaual .. see other constructors)

src/Equinor.Procosys.Preservation.Infrastructure/EntityConfigurations/HistoryConfiguration.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Equinor.Procosys.Preservation.Infrastructure.EntityConfigurations
+{
+    internal class HistoryConfiguration : IEntityTypeConfiguration<History>
+    {
+        public void Configure(EntityTypeBuilder<History> builder)
+        {
+            builder.ConfigurePlant();
+            builder.ConfigureCreationAudit();
+            builder.ConfigureConcurrencyToken();
+
```
janid1967, line 17: PreservationRecordId must configured, make it nullable.

llykkelarsen: Is this necessary? Not required is default and adding the configuration does nothing to the migration script.

janid1967: Need Henning on this one. Should/Can PreservationRecord be ab object in History, or should we just keep PreservationRecordId?
If You see in Step, a Step keep ModeId and ResponsibleId, not Mode or Resposible

src/Equinor.Procosys.Preservation.Domain/AggregateModels/HistoryAggregate/History.cs
```diff
@@ -0,0 +1,67 @@
...
+            EventType eventType
+        ) : base(plant)
+        {
+            Description = description;
+            ObjectId = objectId;
+            ObjectType = objectType;
+            EventType = eventType;
+        }
+
+        public History(
```
hbjorgo, line 31: Is this needed? Isn't it better that the caller does this and that this class is generic? Then we don't need to maintain it with respect to all kinds of types.

src/Equinor.Procosys.Preservation.Infrastructure/EntityConfigurations/HistoryConfiguration.cs
llykkelarsen: @hbjorgo What's your take?

hbjorgo: If any references should be kept, it should be IDs. Because this will be foreign keys to (possible) multiple types of entities, I guess there's no way to set a constraint on the column. It has to be a dangling integer ID. Either we must allow IDs to be dangling (the foreign entity does no longer exist), or we must find a way to add constraints (explicitly have columns for each supported type).

paal123: I think we are mixing the concepts here a little bit @hbjorgo. This discussion is on PreservervationRecord which is just a single nullable reference betweeen History and PreservationRecord tables.

If we only keep the Id and not the navigational property, ti makes it much harder to work with  and query later in my opinion. Especially when we work with IQueryables. In such cases it should have both the id and navigational property in my opinion.

The ObjectId is the dangling id here. Which can in theory be anything, and can never be an navigational property/Object.

src/Equinor.Procosys.Preservation.Domain/AggregateModels/HistoryAggregate/History.cs
paal123: If I understand you correctly here @hbjorgo. You do not want the constructors to do these "checks" and only keep one constructors that accepts whatever is given? Since the validators will do the data quality check?

hbjorgo: The History table might hold items for multiple entities (PreservationRecord, Journeys, etc) in the future. I'm just thinking that if we make it generic, then we don't need to maintain constructors for different types.  The handler that inserts the event will just use the one constructor and insert the needful. We won't break anything by having this, but personally I don't see the need for it.

src/Equinor.Procosys.Preservation.Infrastructure/EntityConfigurations/HistoryConfiguration.cs
hbjorgo: The History table is designed to hold history entries for multiple types of entities. If we include navigation properties, it will be tied to the specific entities. If that's what we want, I'd say we need to make it less generic.

src/Equinor.Procosys.Preservation.Domain/AggregateModels/HistoryAggregate/History.cs
paal123: This is just premature optimization, as we have no indications that this will happen. I suggest we fix it later if we need it.

We have a tight deadline, so I suggest we keep it as it is until we have another need

llykkelarsen: Won't fix cause of tight schedule, but I see what you're saying.

src/Equinor.Procosys.Preservation.Infrastructure/EntityConfigurations/HistoryConfiguration.cs
llykkelarsen: Resolved after group discussion. History will have PreservationRecordId only.

PR #275, opened 2020-06-11
74321 get tag history endpoint

llykkelarsen:


src/tests/Equinor.Procosys.Preservation.Query.Tests/GetHistory/GetHistoryQueryHandlerTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        [TestMethod]
+        public async Task HandleGetHistoryQuery_ShouldReturnCorrectHistory_WhenTagHasHistory()
+        {
+            using (var context = new PreservationContext(_dbContextOptions, _plantProvider, _eventDispatcher,
+                _currentUserProvider))
+            {
+                var dut = new GetHistoryQueryHandler(context);
+                var result = await dut.Handle(_query, default);
+
+                Assert.AreEqual(2, result.Data.Count);
```
janid1967, line 59: Since you already have history domain object and dto, you can create an AssertHistory(History expected, HistoryDto actual) and assert prop for prop in it

janid1967:
We fix Due in later PR's when we know how to create such record with Due

janid1967:
:-)

PR #280, opened 2020-06-16
73944 IsVoided added in TagRequirementsDto

bratli:
IsVoided on Requirement is needed when editing tag.


janid1967:
please add PBI number in front of PR title

PR #281, opened 2020-06-16
74296 preservation tags proxy

janid1967:


src/Equinor.Procosys.Preservation.Query/TagApiQueries/SearchTags/ProcosysTagDto.cs
```diff
@@ -15,7 +15,7 @@ public class ProcosysTagDto
         {
             TagNo = tagNo;
             Description = description;
-            PurchaseOrderNumber = purchaseOrderNumber;
+            PurchaseOrderNumber = purchaseOrderTitle;
```
paal123, line 18: A little bit confusing with Number assignet to tite

janid1967: Can you fix UI if I fix here? Breaking change, therefor no changes from my side.
I guess it hits two places in UI

janid1967: I renamed in new endpoint. Must coordinate with UI to rename another place, I left a ToDo there

PR #285, opened 2020-06-17
74436 Sync Projects and Tags with Main

janid1967:
Command supposed to be called from background job. In this PR its also possible to trigger the command from a normal endpoint in SynchronizeController. Logic in Handler:
* For all projects in preservation where user has access to project (checked via IPermissionCache)
* Get updated project details from Main and update properties we've duplicated into preservation
* For project, get all standard preservation tags in it (we can skip Area tags since the've been created in preservation and never exists in Main)
* Page 50 and 50 for standard tags and get updated tag details from Main and update properties we've duplicated into preservation

We don't check if property has been chaanged since EF Core keep track of changes and don't save if no changes

Removed IProjectRepository.GetAllTagsInProjectAsync(string projectName) and tests. Not in use.

Changes and new tests in BearerTokenApiClient to verify that url not too long

Some refactor in some tests to reuse mock of ICurrentUserProvider and CurrentUserOid

### DIFF
--- a/src/tests/Equinor.Procosys.Preservation.Command.Tests/CommandHandlerTestsBase.cs
+++ b/src/tests/Equinor.Procosys.Preservation.Command.Tests/CommandHandlerTestsBase.cs
@@ -9,6 +9,8 @@ namespace Equinor.Procosys.Preservation.Command.Tests
     [TestClass]
     public abstract class CommandHandlerTestsBase
     {
+        protected readonly Guid CurrentUserOid = new Guid("12345678-1234-1234-1234-123456789123"); // todo rename
+        protected Mock<ICurrentUserProvider> CurrentUserProviderMock;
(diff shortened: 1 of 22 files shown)


src/tests/Equinor.Procosys.Preservation.Command.Tests/CommandHandlerTestsBase.cs
```diff
@@ -9,6 +9,8 @@ namespace Equinor.Procosys.Preservation.Command.Tests
     [TestClass]
     public abstract class CommandHandlerTestsBase
     {
+        protected readonly Guid CurrentUserOid = new Guid("12345678-1234-1234-1234-123456789123"); // todo rename
```
llykkelarsen, line 12: rename/remove todo?

PR #319, opened 2020-07-02
74847 - On-behalf-of token

hbjorgo:
Instead of just forwarding a user's token to the next API when communicating with another API, ask AAD for a token on behalf of the current user. The synchronization service token doesn't represent a user and therefor doesn't support the on-behalf-of flow. It asks for a token specific for the other API, and therefor bypasses this flow.


janid1967:
Please add PBI number to PR title

src/Equinor.Procosys.Preservation.WebApi/Authentication/Authenticator.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            _requestToken = token ?? throw new ArgumentNullException(nameof(token));
+            _canUseOnBehalfOf = isUserToken;
+        }
+
+        public async ValueTask<string> GetBearerTokenOnBehalfOfCurrentUserAsync()
+        {
+            if (_canUseOnBehalfOf)
+            {
+                if (_onBehalfOfUserToken == null || _onBehalfOfUserToken.ExpiresOn < TimeService.UtcNow.AddMinutes(10))
```
Cnordbo, line 32: I dont think this should have a expired check on it. 
If the token is expired, they are unauthenticated to the system, and never hits this endpoint. 

If this doesnt go trough the auth validation first, then just make sure the token isnt expired. Dont add a 10 minute "before it expired" check on it, as that could lead to unwanted results.

src/Equinor.Procosys.Preservation.WebApi/Authentication/Authenticator.cs
```diff
@@ -0,0 +1,73 @@
...
+            }
+            else
+            {
+                return _requestToken;
+            }
+        }
+
+        public async ValueTask<string> GetBearerTokenForApplicationAsync()
+        {
+            if (_applicationToken == null || _applicationToken.ExpiresOn < TimeService.UtcNow.AddMinutes(10))
```
Cnordbo, line 56: I dont think this should have a expired check on it.
If the token is expired, they are unauthenticated to the system, and never hits this endpoint.

If this doesnt go trough the auth validation first, then just make sure the token isnt expired. Dont add a 10 minute "before it expired" check on it, as that could lead to unwanted results.

src/Equinor.Procosys.Preservation.WebApi/appsettings.Development.json
```diff
@@ -9,6 +9,6 @@
   "MigrateDatabase": true,
   "SeedDummyData": false,
   "Synchronization": {
-    "Interval": "0:10:0"
+    "Interval": "0:30:0"
```
Cnordbo, line 12: intentional?

src/Equinor.Procosys.Preservation.WebApi/Authentication/Authenticator.cs
hbjorgo: When a request enters the system, the user is authenticated (the token is verified). If the request then needs to communicate with another API, this kicks in and retrieves another token. It caches this token to avoid asking for a new token multiple times (there's one original request, but it results in multiple requests to the other API). TTL for the token is 60 minutes I believe, so expiration shouldn't be a problem. If so, the whole request would take over 60 minutes, which clearly indicates that there's something wrong.
I'll remove the expiration check.

src/Equinor.Procosys.Preservation.WebApi/Authentication/Authenticator.cs
hbjorgo: Ref. comment above.

src/Equinor.Procosys.Preservation.WebApi/appsettings.Development.json
hbjorgo: Yes.

PR #322, opened 2020-07-03
75119 persist po info on supplier area tag

janid1967:


src/Equinor.Procosys.Preservation.Command/TagCommands/CreateAreaTag/CreateAreaTagCommandHandler.cs
```diff
@@ -114,6 +114,19 @@ private async Task<Tag> CreateAreaTagAsync(CreateAreaTagCommand request)
                 requirements.Add(new TagRequirement(_plantProvider.Plant, requirement.IntervalWeeks, reqDef));
             }
 
+            string purchaseOrderNo = null;
+            string calloff = null;
+
+            if (request.TagType == TagType.PoArea)
+            {
+                var poParts = request.PurchaseOrderCalloffCode.Split('/');
```
hbjorgo, line 122: Do we need a check to see if the split results in an array with length > 0?

janid1967: Yes, should be highlighted when reading unit tests

janid1967: In selector in UI, either PO or PO/CO can be chosen

janid1967: What I should do, is to throw excdeption if PurchaseOrderCalloffCode is null or empty

PR #337, opened 2020-07-14
GetPreservationRecords extended dto

bratli:


src/Equinor.Procosys.Preservation.Query/GetHistory/GetHistoryQueryHandler.cs
```diff
@@ -23,7 +24,10 @@ public async Task<Result<List<HistoryDto>>> Handle(GetHistoryQuery request, Canc
             var tagHistory = await (from h in _context.QuerySet<History>()
```
bratli, line 24: @kfaln

PR #340, opened 2020-07-21
75278: Delete requirement definition

bratli:
Delete requirement definition.

Have deleteRequirementDefinition in RequirementTypeRepository because using remove as on DeleteStep does not delete object from database (just removes parentId). (ParentId on requirementDefinition is not nullable in database, so would have to change this if we wanted to remove requirement definition in the same way as we are "removing" step for example). 

Need to create composite key to achieve a complete delete, and can then remove this remove method from 

RequirementTypeRepository. The composite key for Requirement Definition would be Id and RequirementTypeId.

### DIFF
--- a/src/tests/Equinor.Procosys.Preservation.Command.Tests/RequirementTypeCommands/DeleteRequirementDefinition/DeleteRequirementDefinitionCommandValidatorTests.cs
+++ b/src/tests/Equinor.Procosys.Preservation.Command.Tests/RequirementTypeCommands/DeleteRequirementDefinition/DeleteRequirementDefinitionCommandValidatorTests.cs
@@ -1,0 +1,90 @@
+﻿using System.Threading.Tasks;
+using Equinor.Procosys.Preservation.Command.RequirementTypeCommands.DeleteRequirementDefinition;
+using Equinor.Procosys.Preservation.Command.Validators.RequirementDefinitionValidators;
+using Equinor.Procosys.Preservation.Command.Validators.RequirementTypeValidators;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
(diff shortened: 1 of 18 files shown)


src/tests/Equinor.Procosys.Preservation.Command.Tests/RequirementTypeCommands/DeleteRequirementDefinition/DeleteRequirementDefinitionCommandValidatorTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            _requirementDefinitionValidatorMock.Setup(r => r.TagFunctionRequirementsExistAsync(_requirementDefinitionId, default)).Returns(Task.FromResult(true));
+
+            var result = _dut.Validate(_command);
+
+            Assert.IsFalse(result.IsValid);
+            Assert.AreEqual(1, result.Errors.Count);
+            Assert.IsTrue(result.Errors[0].ErrorMessage.StartsWith("Tag function requirement with this requirement definition exists!"));
+        }
+    }
+}
```
llykkelarsen, line 90: add Validate_ShouldFail_WhenInvalidRowVersion()
        {   
           const string invalidRowVersion = "String";
           var command = new DeleteRequirementDefinitionCommand(_tagId, _actionId, invalidRowVersion);
           _rowVersionValidatorMock.Setup(r => r.IsValid(invalidRowVersion, default)).Returns(Task.FromResult(false));

           var result = _dut.Validate(command);

            Assert.IsFalse(result.IsValid);
            Assert.AreEqual(1, result.Errors.Count);
            Assert.IsTrue(result.Errors[0].ErrorMessage.StartsWith("Not a valid RowVersion!"));
        }

PR #361, opened 2020-08-12
76012 delete saved filter endpoint

llykkelarsen:


src/Equinor.Procosys.Preservation.Infrastructure/Repositories/PersonRepository.cs
```diff
@@ -7,10 +7,10 @@ namespace Equinor.Procosys.Preservation.Infrastructure.Repositories
 {
     public class PersonRepository : RepositoryBase<Person>, IPersonRepository
     {
+        private readonly PreservationContext _context;
```
janid1967, line 10: move PreservationContext _context; to baseclass and init via constructor. Expand 3-param constructor of RepositoryBase with PreservationContext as first param. We need it allover the place regarding delete. We already have another repo with PreservationContext _context; as private which then can be rewritten in same manner

llykkelarsen: not 100% sure I got it right, can you check my latest commit? @janid1967

janid1967: @llykkelarsen Did you see comment from me 2h ago?

PR #371, opened 2020-08-20
Minor fixes

hbjorgo:
Some minor fixes.


janid1967 (changes requested):
The change in CacheManager lead to 10 red tests. Guess tests misses setup of TimeService

janid1967 (approved):

PR #383, opened 2020-08-27
75233 error message when uploading big files

janid1967:


src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/UploadBaseDtoValidator.cs
```diff
@@ -27,16 +27,22 @@ public UploadBaseDtoValidator(IOptionsMonitor<AttachmentOptions> options)
...
-                .WithMessage($"Maximum file size is {options.CurrentValue.MaxSizeKb}kB!");
+                .WithMessage($"Maximum file size is {options.CurrentValue.MaxSizeMb}MB!");
 
             bool BeAValidFile(string fileName)
             {
                 var suffix = Path.GetExtension(fileName?.ToLower());
                 return suffix != null && options.CurrentValue.ValidFileSuffixes.Contains(suffix) && fileName?.IndexOfAny(Path.GetInvalidFileNameChars()) == -1;
             }
+            
+            bool NotBeTooBigFile(long fileSizeInBytes)
```
llykkelarsen, line 41: Got stuck on this method name trying to figure out whether too big was supposed to return true or false.. could we lose the 'Not'? F.ex BeAppropriateFileSize or BeSmallerThanMaxSize?

llykkelarsen (approved):
Approved with suggestion

PR #388, opened 2020-08-28
BUGFIX how validators check if a child element exists

janid1967:


janid1967:
@llykkelarsen I need to change this, not correct yet

PR #408, opened 2020-09-18
78420 Try fixing 413 Request Entity Too Large Errors

janid1967:
Work in localhost. Need to test in real Azure environment

See also comment in BUG 78420 - we probably need to adjust nginx.ingress.kubernetes.io/proxy-body-size as well

### DIFF
--- a/src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagsController.cs
+++ b/src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagsController.cs
@@ -266,6 +266,7 @@ namespace Equinor.Procosys.Preservation.WebApi.Controllers.Tags
             return Redirect(result.Data.ToString());
         }
 
+        [DisableRequestSizeLimit]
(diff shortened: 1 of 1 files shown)


src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagsController.cs
```diff
@@ -266,6 +266,7 @@ public TagsController(IMediator mediator, IExcelConverter excelConverter)
             return Redirect(result.Data.ToString());
         }
 
+        [DisableRequestSizeLimit]
```
paal123, line 269: I think there might be a DOS risk related to removing this limit. But the risk is probably rather small

paal123 (approved):
With comments

src/Equinor.Procosys.Preservation.WebApi/Controllers/Tags/TagsController.cs
janid1967: wo will not merge this one

PR #412, opened 2020-09-24
Handling users without firstname and lastname

paal123:
Doing the same "magic" as ad sync when handling affiliate user without first name and last name


llykkelarsen (approved):

kfaln (approved):

janid1967 (approved):

src/Equinor.Procosys.Preservation.WebApi/Middleware/VerifyOidInDbMiddleware.cs
```diff
@@ -27,6 +27,30 @@ public class VerifyOidInDbMiddleware
                 var givenName = httpContextUser.Claims.TryGetGivenName();
                 var surName = httpContextUser.Claims.TryGetSurName();
 
+                if (string.IsNullOrEmpty(givenName) && string.IsNullOrEmpty(surName))
+                {
+                    var fullName = httpContextUser.Claims.TryGetFullName();
+
```
janid1967, line 33: Would be nice to check if fullname is null and throw exception explaining the problem. Better than null-pointer exception

PR #423, opened 2020-10-08
78422 Add change of journey in history

janid1967:
see commit comments

### DIFF
--- a/src/Equinor.Procosys.Preservation.Command/Validators/TagValidators/TagValidator.cs
+++ b/src/Equinor.Procosys.Preservation.Command/Validators/TagValidators/TagValidator.cs
@@ -44,6 +44,12 @@ namespace Equinor.Procosys.Preservation.Command.Validators.TagValidators
             return tag != null && tag.Status == status;
         }
 
+        public async Task<bool> VerifyTagTypeAsync(int tagId, TagType tagType, CancellationToken token)
+        {
+            var tag = await GetTagWithoutIncludes(tagId, token);
+            return tag != null && tag.TagType == tagType;
+        }
+
         public async Task<bool> HasANonVoidedRequirementAsync(int tagId, CancellationToken token)
         {
             var tag = await GetTagWithRequirements(tagId, token);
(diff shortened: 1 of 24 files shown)


src/Equinor.Procosys.Preservation.Command/Validators/TagValidators/TagValidator.cs
```diff
@@ -44,6 +44,12 @@ public async Task<bool> VerifyPreservationStatusAsync(int tagId, PreservationSta
             return tag != null && tag.Status == status;
         }
 
+        public async Task<bool> VerifyTagTypeAsync(int tagId, TagType tagType, CancellationToken token)
+        {
+            var tag = await GetTagWithoutIncludes(tagId, token);
```
bratli, line 49: Just a note, but I think GetTagWithoutIncludes could be renamed to something better (or maybe I'm just not familiar enough with the code). Without Includes what ? Without includes requirements ? Or is it get tag by ID?

bratli (approved):

src/Equinor.Procosys.Preservation.Command/Validators/TagValidators/TagValidator.cs
janid1967: What about GetTagWithoutChilds?  GetBareTag? GetJustTag? GetTagSimple?
The method return a tag without any other data "under" (as Requirements, Attachchments, Actions)

bratli: Maybe getTagSimple ? But it is fine as it is as well, because if you were to get "children" then you would write .Includes(child), right?

janid1967: Yes, that's the default behavior; Get the tag with all its children. I guess you've heard we've talking about the "DefaultQuery" ... that is with all includes, which in some cases will be a massive query from DB

PR #427, opened 2020-10-09
78680 Use new project features from main api

janid1967:
Part 2 of #425
The goal for this change is to distinguish between accessing a invalid project vs a project where user don't have access.

* Refactor IPermissionApiService -> GetAllProjectsAsync now get all projects, also those where user don't have access. New property ProcosysProject.HasAccess tell if user has access
* Refactored CurrentPlantMiddleware. Throw exception, not return BadRequest, if headers can't be found. This **IS** unexcepted. 
* New behavior in MediatR pipeline: If request is requesting data froma project, check if that project is a valid one. Return BadRequest if not valid. 
* HttpContextExtension which write a correct BadRequest json back to client. Also used in PlantValidatorMiddleware made in prev PR
* NB! Sequence of behaviors changed. Validate query/command before checking access and if project is valid. This because we want a BadRequest before Forbidden


Cnordbo (approved):
Are we missing some tests for unhappy paths?

janid1967:
> Are we missing some tests for unhappy paths?
Should be covered by future integration tests (end-to-end)

PR #448, opened 2020-10-30
78684 Integration tests

janid1967:
Ideas in code:

**Scaffolding**
* Uses WebAppliactionFactory to bootstrap the web application in memory for functional end to end tests 
* The TestFactory, inheriting WebAppliactionFactory, is initiated for each test run. (static instance). 
* The TestFactory creates testUsers with HttpClients to be used for invoking endpoints in tests
* The TestFactory forward all requests to a custom IntegrationTestAuthHandler which creates the needed Bearer token
* The TestFactory creates all HttpClients at startup. (Expensive startup due to this!)
* The TestFactory uses Mock to mock requesting ProCoSys for plants, projects and permissions 
* The mock of plants returns 2 plants, one where users have access, and one without access
* The mock of projects returns 2 projects, one where users have access, and one without access
* All users (minus Anonymous and Hacker) has access to one plant, which is used in positive tests
* The mock of permissions differs for each user
* The Hacker user simulates an authenticated user, but lack of any plant, project or permission
* The TestFactory uses LocalDB for database. 
* The TestFactory created the database, inclusive the actual .mdf file.
* The TestFactory run all database migrations when creating the database, meaning schema is created from scratch each time (No use of Respawn NuGet needed for now)
* The database (incl .mdf) is deleted at startup if exists.
*  he database (incl .mdf) is deleted in teardown (Dispose). 
* The Docker compose Job do not run tests now. Docker uses ubuntu, and LocalDB is not supported under Linux. 
* Pipeline is modified (by Terje G) with a new Job, running both Unit- and Integration tests before the Docker Build job

**Tests**
* A test should not have main focus on business logic. This should be covered in Unit tests
* The tests should cover that endpoints are secured by correct RoleClaim (i.e. ProCoSys permission as LIBRARY_PRESERVATION/CREATE)
* A test does not have to test access to correct project, this is already covered in Unit tests
* Each test is responsible for creating needed test data. (i.e. seed its data by using endpoints)
* A test does not depend on result from another test (sequence of test should not be significant)
* The tests for a controller should cover positive tests for CRUD
* The tests should cover use of RowVersion and assert that a PUT updates the RowVersion (not covered in Unit tests by now)


Cnordbo (approved):

src/tests/Equinor.Procosys.Preservation.WebApi.IntegrationTests/TestFactory.cs
```diff
@@ -0,0 +1,380 @@
...
+            return spScope.ServiceProvider.GetRequiredService<PreservationContext>();
+        }
+
+        private string GetTestDbConnectionString(string projectDir)
+        {
+            var dbName = "IntegrationTestsDB";
+            var dbPath = Path.Combine(projectDir, $"{dbName}.mdf");
+            
+            // Set Initial Catalog to be able to delete database!
+            return $"Server=(LocalDB)\\MSSQLLocalDB;Initial Catalog={dbName};Integrated Security=true;AttachDbFileName={dbPath}";
```
Cnordbo, line 175: Should this be in appsettings?

src/tests/Equinor.Procosys.Preservation.WebApi.IntegrationTests/TestFactory.cs
```diff
@@ -0,0 +1,380 @@
...
+            if (descriptor != null)
+            {
+                services.Remove(descriptor);
+            }
+
+            services.AddDbContext<PreservationContext>(context => context.UseSqlServer(_connectionString));
+
+            using var serviceProvider = services.BuildServiceProvider();
+
+            using var scope = serviceProvider.CreateScope();
```
Cnordbo, line 130: unfamiliar with this kind of `using` statement without scoping. 

Expected: 
```c#
using (var scope = serviceProvider.CreateScope()) {
--- Executing code
}
```

Cnordbo (approved):

src/tests/Equinor.Procosys.Preservation.WebApi.IntegrationTests/TestFactory.cs
janid1967: > unfamiliar with this kind of `using` statement without scoping.
> 
> Expected:
> 
> ```cs
> using (var scope = serviceProvider.CreateScope()) {
> --- Executing code
> }
> ```

C# 8.0
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement

Won't fix

src/tests/Equinor.Procosys.Preservation.WebApi.IntegrationTests/TestFactory.cs
janid1967: I don't think so.
The code decide where file should be created. The path is dynamic. 

OK with with Won't fix  @Cnordbo ?

src/Equinor.Procosys.Preservation.WebApi/Startup.cs
```diff
@@ -26,6 +26,7 @@ namespace Equinor.Procosys.Preservation.WebApi
 {
     public class Startup
     {
+        public const string IntegrationTestEnvironment = "IntegrationTests";
```
hbjorgo, line 29: Create extension method instead of using the string comaprison. (IsIntegrationTest())

src/tests/Equinor.Procosys.Preservation.WebApi.IntegrationTests/TestFactory.cs
Cnordbo: Based on the current spec of the connection string, that it doesnt support relative URIs and you need a string replacement anyway, this can be left as is.

src/tests/Equinor.Procosys.Preservation.WebApi.IntegrationTests/IntegrationTestAuthHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+using System.Threading.Tasks;
+using Equinor.Procosys.Preservation.WebApi.Authorizations;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+
+namespace Equinor.Procosys.Preservation.WebApi.IntegrationTests
+{
+    internal class IntegrationTestAuthHandler : AuthenticationHandler<IntegrationTestAuthOptions>
```
hbjorgo, line 16: Any specific reason for this naming convention? Current naming convention would name it AuthHandlerIntegrationTests.

src/Equinor.Procosys.Preservation.WebApi/Startup.cs
janid1967: No comparison needed. This is the const

hbjorgo (approved):

PR #455, opened 2020-11-16
77899 how validator check when children exists - Vol 1

janid1967:
See commits

This is first PR of several to fix same problem in many validators. See 77899 for documentation of which validators to be fixed. This PR fix DeleteActionAttachmentCommandValidator and DeleteTagAttachmentCommandValidator


src/Equinor.Procosys.Preservation.Command/ActionAttachmentCommands/Delete/DeleteActionAttachmentCommandValidator.cs
```diff
@@ -23,31 +21,23 @@ public class DeleteActionAttachmentCommandValidator : AbstractValidator<DeleteAc
             RuleFor(command => command)
                 .MustAsync((command, token) => NotBeAClosedProjectForTagAsync(command.TagId, token))
                 .WithMessage(command => $"Project for tag is closed! Tag={command.TagId}")
-                .MustAsync((command, token) => BeAnExistingTagAsync(command.TagId, token))
-                .WithMessage(command => $"Tag doesn't exist! Tag={command.TagId}")
+                .MustAsync(BeAnExistingActionAttachmentAsync)
```
llykkelarsen, line 24: Does no parameters mean that it automatically sends the entire command and token? Why don't we pass tagId, actionId, attachmentId here?

llykkelarsen (approved):

src/Equinor.Procosys.Preservation.Command/ActionAttachmentCommands/Delete/DeleteActionAttachmentCommandValidator.cs
janid1967: A)                .MustAsync((command, token) => BeAnExistingActionAsync(command, token))
is equal to
B)                .MustAsync(BeAnExistingActionAsync)

since (command, token) match exact the input paramlist for BeAnExistingActionAttachmentAsync. ReSharper suggest to use method group B instead of lambda expression A. And yes, we could pass tagId, actionId, attachmentId as you suggest. I preferred method group, more easy to read

OK?

llykkelarsen: OK 👍

PR #472, opened 2021-01-18
80421 details in history in excel bugfix

janid1967:
Bugfix 1 when creating Excel with details: 
Need to handle that FieldValues for all Fields in a Requirement Def don't have to exists for all preservation periods since a field can be added to a Requirement Def after preservation periods are made

Bugfix 2: GetPreservationRecordQueryHandler - The interval was taken from the Requirement Definition, not from the actual TagRequirement (showed wrong in UI)

### DIFF
--- a/src/tests/Equinor.Procosys.Preservation.Query.Tests/GetPreservationRecord/GetPreservationRecordQueryHandlerTests.cs
+++ b/src/tests/Equinor.Procosys.Preservation.Query.Tests/GetPreservationRecord/GetPreservationRecordQueryHandlerTests.cs
@@ -24,6 +24,8 @@ namespace Equinor.Procosys.Preservation.Query.Tests.GetPreservationRecord
         private RequirementType _requirementType;
         private RequirementDefinition _requirementDefinitionWithoutOneField;
         private Field _infoField;
+        private int _interval;
+        private string _comment = "comment";
 
         protected override void SetupNewDatabase(DbContextOptions<PreservationContext> dbContextOptions)
         {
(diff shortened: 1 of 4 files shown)


src/tests/Equinor.Procosys.Preservation.Query.Tests/GetPreservationRecord/GetPreservationRecordQueryHandlerTests.cs
```diff
@@ -34,9 +36,10 @@ protected override void SetupNewDatabase(DbContextOptions<PreservationContext> d
                 _requirementType = AddRequirementTypeWith1DefWithoutField(context, "RT", "RD", RequirementTypeIcon.Other, 1);
                 _requirementDefinitionWithoutOneField =
                     _requirementType.RequirementDefinitions.Single();
+                _interval = _requirementDefinitionWithoutOneField.DefaultIntervalWeeks * 2;
```
bratli, line 39: I dont know the logic behind this, but it is correct to multiply by 2? (or is it just "random" for testing purpose?)

bratli (approved):

src/tests/Equinor.Procosys.Preservation.Query.Tests/GetPreservationRecord/GetPreservationRecordQueryHandlerTests.cs
janid1967: random, and to be sure that interval used in TagRequirement differ from _requirementDefinitionWithoutOneField.DefaultIntervalWeeks. If interval in TagRequirement accidently becomes equal to_requirementDefinitionWithoutOneField.DefaultIntervalWeeks, the new Assert at line 92 are not "secure".
I can add a comment on it

PR #524, opened 2021-04-15
Time service

hbjorgo:
Using TimeService NuGet package.


paal123:
Not in accordance with Equinor policy

PR #554, opened 2021-06-10
82463 trace more info regarding cross plant access check

janid1967:
[AB#82463](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/82463)

### DIFF
--- a/src/Equinor.ProCoSys.Preservation.WebApi/Authorizations/AuthorizationOptions.cs
+++ b/src/Equinor.ProCoSys.Preservation.WebApi/Authorizations/AuthorizationOptions.cs
@@ -26,7 +26,7 @@ namespace Equinor.ProCoSys.Preservation.WebApi.Authorizations
 
         private void Transform()
         {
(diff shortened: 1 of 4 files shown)


src/Equinor.ProCoSys.Preservation.WebApi/Authorizations/AuthorizationOptions.cs
```diff
@@ -1,29 +1,42 @@
...
-                return userOids;
+                _crossPlantUserOidList = value;
+                Transform();
             }
+        }
 
-            foreach (var oid in CrossPlantUserOidList.Split(","))
+        private void Transform()
+        {
+            if (_crossPlantUserOidList == null)
```
Cnordbo, line 29: should probably be `if(string.nullOrEmpty(_crossPlantUserOidList))`

Cnordbo (approved):

PR #588, opened 2021-08-02
83491 Voided tags active after a week bugfix

llykkelarsen:
[AB#83491](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/83491)

Bug was: Any tag which had been voided showed up in the unvoided list a week later. 

Solution: Sync bus wrongfully set tag isvoided, this was not done in the prior sync solution. 'Voiding' a tag in preservation is not the same thing as voiding a tag in main, so this property will be ignored in preservation.

### DIFF
--- a/src/tests/Equinor.ProCoSys.Preservation.WebApi.Tests/Synchronization/BusReceiverServiceTests.cs
+++ b/src/tests/Equinor.ProCoSys.Preservation.WebApi.Tests/Synchronization/BusReceiverServiceTests.cs
@@ -600,7 +600,6 @@ namespace Equinor.ProCoSys.Preservation.WebApi.Tests.Synchronization
             // Act
             await _dut.ProcessMessageAsync(PcsTopic.Tag, message, new CancellationToken(false));
 
-
             // Assert
             Assert.AreEqual(area, _tag1.AreaCode);
             Assert.AreEqual(areaDescription, _tag1.AreaDescription);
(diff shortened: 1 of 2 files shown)


src/tests/Equinor.ProCoSys.Preservation.WebApi.Tests/Synchronization/BusReceiverServiceTests.cs
```diff
@@ -638,10 +636,30 @@ public async Task HandleTagTopic_ShouldUpdateTagAndChangeTagNo()
...
+            var area = "Area69";
+            var areaDescription = "Area69 Description";
+            var discipline = "ABC";
+            var disciplineDescription = "ABC Desc";
+            var callOffNo = "123";
+            var poNo = "321";
+            var tagFunctionCodeNew = "FCS123";
+            var message =
+                $"{{\"TagNo\" : \"{TagNo1}\",\"Description\" : \"Test 123\",\"ProjectName\" : \"{Project1Name}\",\"McPkgNo\" : \"{McPkg1}\",\"CommPkgNo\" : \"{CommPkg1}\",\"AreaCode\" : \"{area}\",\"AreaDescription\" : \"{areaDescription}\",\"DisciplineCode\" : \"{discipline}\",\"DisciplineDescription\" : \"{disciplineDescription}\",\"CallOffNo\" : \"{callOffNo}\",\"PurchaseOrderNo\" : \"{poNo}\",\"TagFunctionCode\" : \"{tagFunctionCodeNew}\",\"IsVoided\" : true,\"Plant\" : \"{Plant}\"}}";
+
```
paal123, line 655: I would suggest an assert before the act as well, to verify that the value is not affected.

paal123 (approved):
With suggestions.
Remember to also link DevOps number

PR #624, opened 2021-10-19
Feature/85891   adding config map

paal123:
Finalizing leader pod functionality.
All leader pod functionality is located in the service bus package. 
Only config here i Preservation to enable it.

[AB#85891](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/85891)


paal123:
@janid1967 Please re-review

PR #664, opened 2021-12-06
Dotnet6

TordJoranger:


paal123:
.Net 6 is not supported in VS 2019. We must make sure all developer has updated before completing this PR.

janid1967:
Also failing tests

PR #665, opened 2021-12-07
Refresh app configuration when sentinel value changes

chormaz:
The app configuration settings are now updated 30 secs. after a change in the sentinel value is done in Azure App Configuration.

### DIFF
--- a/src/Equinor.ProCoSys.Preservation.WebApi/Authentication/Authenticator.cs
+++ b/src/Equinor.ProCoSys.Preservation.WebApi/Authentication/Authenticator.cs
@@ -11,18 +11,18 @@ namespace Equinor.ProCoSys.Preservation.WebApi.Authentication
 {
     public class Authenticator : IBearerTokenProvider, IBearerTokenSetter, IAuthenticator
     {
-        private readonly IOptionsSnapshot<AuthenticatorOptions> _options;
+        private readonly IOptionsMonitor<AuthenticatorOptions> _options;
         private readonly ILogger<Authenticator> _logger;
         private string _requestToken;
         private string _onBehalfOfUserToken;
         private string _applicationToken;
         private readonly string _secretInfo;
 
-        public Authenticator(IOptionsSnapshot<AuthenticatorOptions> options, ILogger<Authenticator> logger)
+        public Authenticator(IOptionsMonitor<AuthenticatorOptions> options, ILogger<Authenticator> logger)
         {
             _options = options;
             _logger = logger;
-            var apiSecret = _options.Value.PreservationApiSecret;
+            var apiSecret = _options.CurrentValue.PreservationApiSecret;
             _secretInfo = $"{apiSecret.Substring(0, 2)}***{apiSecret.Substring(apiSecret.Length - 1, 1)}";
             AuthenticationType = AuthenticationType.OnBehalfOf;
         }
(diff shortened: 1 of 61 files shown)


janid1967:
Remove my test-configs named Authenticator:GlobalSetting and Authenticator:ScopedSetting. Remove both from code + tidy in configs pcs-pres-non-prod-config / pcs-pres-prod-config

janid1967 (changes requested):
Remove my test-configs named Authenticator:GlobalSetting and Authenticator:ScopedSetting. Remove both from code + tidy in configs pcs-pres-non-prod-config / pcs-pres-prod-config

src/Equinor.ProCoSys.Preservation.WebApi/Authentication/Authenticator.cs
```diff
@@ -18,7 +18,7 @@ public class Authenticator : IBearerTokenProvider, IBearerTokenSetter, IAuthenti
         private string _applicationToken;
         private readonly string _secretInfo;
 
-        public Authenticator(IOptions<AuthenticatorOptions> options, ILogger<Authenticator> logger)
+        public Authenticator(IOptionsSnapshot<AuthenticatorOptions> options, ILogger<Authenticator> logger)
```
janid1967, line 21: Why this?

src/Equinor.ProCoSys.Preservation.WebApi/Program.cs
```diff
@@ -17,38 +17,39 @@ public static void Main(string[] args)
...
+                                 .ConfigureKeyVault(kv =>
+                                 {
+                                     kv.SetCredential(new DefaultAzureCredential());
+                                 })
+                                 .Select(KeyFilter.Any)
+                                 .Select(KeyFilter.Any, context.HostingEnvironment.EnvironmentName)
+                                 .ConfigureRefresh(refreshOptions =>
+                                 {
+                                     refreshOptions.Register("Sentinel", true);
+                                     refreshOptions.SetCacheExpiration(TimeSpan.FromSeconds(30));
```
janid1967, line 41: Revert back to 5 minutes.
It's a resaon to use 5 minutes rather than 30 sec. I'll explain the scenario later

src/Equinor.ProCoSys.Preservation.WebApi/Authentication/Authenticator.cs
chormaz: image

janid1967: As mentioned yesterday: Then all use of IOptionMonitor need to be replaced

src/Equinor.ProCoSys.Preservation.WebApi/Middleware/VerifyPreservationApiClientExists.cs
```diff
@@ -15,12 +15,12 @@ namespace Equinor.ProCoSys.Preservation.WebApi.Middleware
     public class VerifyPreservationApiClientExists : IHostedService
     {
         private readonly IServiceScopeFactory _serviceProvider;
-        private readonly IOptionsSnapshot<AuthenticatorOptions> _options;
+        private readonly IOptionsMonitor<AuthenticatorOptions> _options;
```
janid1967, line 18: Yet another type of IOptions-stuff .. so why this type needed just here, and no other place where you started to use IOptionsSnapshot

chormaz: We can't use IOptionsSnapshot for Hosted Services. The difference between the two is that IOptionsSnapshot keeps the value through out the execution of a request whilst IOptionsMonitor can change the current value during the execution of a request. The chance for it to cause real-life errors is close to zero. Therefore the change of type.

janid1967 (approved):

PR #666, opened 2021-12-09
88458 - Refactored test setup so Journey tests don't mess up journey used in Tag tests

janid1967:
AB#88458


chormaz (approved):
Testene kjører uten feil nå.

PR #674, opened 2021-12-29
88576 optimistic concurrency integration tests

janid1967:
[AB#88576](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/88576)

Near 100% test coverage now, with focus testing Conflict when using wrong RowVersion

**Found a BUG which is fixed**: Validators for TagFunction need to honor both TagFunctionCode + RegisterCode when checking for exists / isvoided. Old code crash when Voiding/Unvoiding a TagFunctionCode/RegisterCode when same TagFunctionCode exists in multiple registers (due to a SingleOrDefault)


janid1967:
Need review from somebody @TordJoranger @terjewoldseth @danhalvorsen

paal123 (approved):

PR #819, opened 2022-12-01
98387 procosysguid

janid1967:
[AB#98387](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/98387)


src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/ProjectAggregate/Tag.cs
```diff
@@ -39,6 +39,7 @@ protected Tag()
         public Tag(
             string plant,
             TagType tagType,
+            Guid? proCoSysGuid,
```
TordJoranger, line 42: is this a 2 step implementation where guid is non nullable later, or can it be null situationally? If first, maybe a comment?

TordJoranger: Question answered in next file, disregard :D

src/Equinor.ProCoSys.Preservation.WebApi/Seeding/Seeders.cs
```diff
@@ -93,14 +93,17 @@ public static void AddTags(this IEnumerable<Project> projects, int entryCountPrP
                     var tag = new Tag(
                         plant, 
                         TagType.Standard,
+                        Guid.NewGuid(),
```
TordJoranger, line 96: not a test project, but I assume this is for test purposes of some sort.

TordJoranger (approved):

src/Equinor.ProCoSys.Preservation.WebApi/Seeding/Seeders.cs
janid1967: As name of class say: Used for seed dummy data into database. Used early in project, and has not been used since then. (We could consider to throw away code, but it don't disturb that much either)

PR #848, opened 2022-12-12
98517 - Temporary topics with string as ProCoSysGuids

janid1967:
[AB#98517](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/98517)


TordJoranger (approved):
jeg skriver mine som public Entity {get; set;} på en linje, men you do you

PR #850, opened 2022-12-22
98580 - Change rule when a Journey can be deleted

janid1967:
[AB#98580](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/98580)
Now a Journey can be deleted if it has steps, but no step in Journey can't have a tag

### DIFF
--- a/src/tests/Equinor.ProCoSys.Preservation.Infrastructure.Tests/Repositories/JourneyRepositoryTests.cs
+++ b/src/tests/Equinor.ProCoSys.Preservation.Infrastructure.Tests/Repositories/JourneyRepositoryTests.cs
@@ -18,12 +18,15 @@ namespace Equinor.ProCoSys.Preservation.Infrastructure.Tests.Repositories
     public class JourneyRepositoryTests : RepositoryTestBase
     {
         private const int JourneyId = 5;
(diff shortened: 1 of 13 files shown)


src/tests/Equinor.ProCoSys.Preservation.Infrastructure.Tests/Repositories/JourneyRepositoryTests.cs
```diff
@@ -106,7 +121,7 @@ public async Task GetJourneysWithAutoTransferStepsAsync_MatchingAutoTransferMeth
         }
 
         [TestMethod]
-        public void Remove_WhenJourneyIsVoided_ShouldCallRemoveForJourney()
+        public void Remove_WhenJourneyIsVoided_ShouldRemoveJourneyAndStepsFromContext()
```
bratli, line 124: Maybe add a test that we cannot remove journey that has steps that has tags?

bratli: Here and/or integration test

bratli (approved):

PR #863, opened 2023-01-03
98716 - even more logging

janid1967:
[AB#98716](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/98716)


krsteen (approved):
Her skal det logges! :)

PR #896, opened 2023-03-08
Tech/96322 nuget updates

chormaz:
Updated all nuget packages to latest version. 
Also migrated all projects to .net7.
Excel-generation needs to be tested.

[AB#96322](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/96322)


janid1967 (changes requested):
We must keep 6.0 and have a plan upgrading to 7.0

chormaz:
Made changes to a new branch.

PR #897, opened 2023-03-09
100144 - Use Common NuGets

janid1967:
[AB#100144](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/100144)


src/Equinor.ProCoSys.Preservation.Domain/TagOptions.cs
```diff
@@ -3,5 +3,6 @@
     public class TagOptions
     {
         public int IsNewHours { get; set; } = 48;
+        public int TagSearchPageSize { get; set; } = 10000;
```
bratli, line 6: Is this the same that you showed me that would be added to config in Azure? Why do we need it here as well if that is the case?

src/Equinor.ProCoSys.Preservation.MainApi/Equinor.ProCoSys.Preservation.MainApi.csproj
```diff
@@ -5,9 +5,14 @@
...
   <ItemGroup>
+    <PackageReference Include="Equinor.ProCoSys.Auth" Version="0.1.1" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="6.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.2" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.3" />
     <PackageReference Include="System.Text.Json" Version="6.0.6" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\Equinor.ProCoSys.Preservation.Domain\Equinor.ProCoSys.Preservation.Domain.csproj" />
```
bratli, line 15: Preservation.MainApi will eventually be deleted? Just remember Henning saying to be "carefull" adding project references

src/tests/Equinor.ProCoSys.Preservation.WebApi.IntegrationTests/RequirementTypes/RequirementTypesControllerNegativeTests.cs
```diff
@@ -511,12 +518,13 @@ public async Task CreateRequirementDefinition_AsAnonymous_ShouldReturnUnauthoriz
...
         [TestMethod]
-        public async Task CreateRequirementDefinition_AsHacker_ShouldReturnForbidden_WhenUnknownPlant()
+        public async Task CreateRequirementDefinition_AsHacker_ShouldReturnBadRequest_WhenUnknownPlant()
             => await RequirementTypesControllerTestsHelper.CreateRequirementDefinitionAsync(
                 UserType.Hacker, TestFactory.UnknownPlant,
                 ReqTypeAIdUnderTest,
                 Guid.NewGuid().ToString(),
-                expectedStatusCode:HttpStatusCode.Forbidden);
+                expectedStatusCode: HttpStatusCode.BadRequest,
+                expectedMessageOnBadRequest: "is not a valid plant");
```
bratli, line 527: Do we want this? This gives the hacker information that we might not want him to have

src/Equinor.ProCoSys.Preservation.Domain/TagOptions.cs
janid1967: The default value for the old config was much less than 10000. (I think it was 100)
When I now move this config to a new class, I default it as the value we found was smart to use in Azure config
This also mean that in this version, we don't have to setup the TagOptions.TagSearchPageSize in Azure config (before we want to change it to other than 10000)

src/Equinor.ProCoSys.Preservation.MainApi/Equinor.ProCoSys.Preservation.MainApi.csproj
janid1967: MainApi will not be deleted. 
 Reason for adding Domain to MainApi (and other) projects is the new TagOptions.cs class, needed cross projects

bratli (approved):

PR #899, opened 2023-03-13
Tech/96322 - Updated packages to latest version except those dependent on .net7.0.

chormaz:
Updated all packages used to latest version except those dependent on .net7.0 (see image below).
Also embedded google font Carlito due to update of ClosedXML - this caused an error in non-windows environments due to required Calibri-font. Now using Carlito as fallback should fix the issue with just minor differences in appereance.
This solution is adviced here: https://closedxml.readthedocs.io/en/latest/tips/missing-font.html
Excel generation should be tested.

[AB#96322](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/96322)
image


src/Equinor.ProCoSys.Preservation.WebApi/Excel/ExcelConverter.cs
```diff
@@ -94,7 +96,15 @@ public static class HistorySheetColumns
         private readonly ILogger<ExcelConverter> _logger;
         private Domain.Time.Timer _timer;
 
-        public ExcelConverter(ILogger<ExcelConverter> logger) => _logger = logger;
+        public ExcelConverter(ILogger<ExcelConverter> logger)
+        {
+            _logger = logger;
+            using (var fallbackFontStream = Assembly.GetExecutingAssembly()
```
janid1967, line 102: nice work!

janid1967 (approved):
Approved, but we merge it to master after testing it in a docker env as agrred

PR #918, opened 2023-04-19
102031 - Allow transfer when in service

janid1967:
[AB#102031](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/102031)


krsteen (approved):
Gudd! Likte enkapsulering av GetNextStep(StepId) != null inn i HasNextStep(stepId)! Mye mer behagelig å lese enn de der != null over alt

PR #936, opened 2023-06-07
102986 - Also migrate History records where new SourceGuid is still null guid

janid1967:
AB#102986


src/Equinor.ProCoSys.Preservation.Infrastructure/Migrations/20230607071458_MigrateNullGuidsInHistory.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Equinor.ProCoSys.Preservation.Infrastructure.Migrations
+{
+    /// <inheritdoc />
+    public partial class MigrateNullGuidsInHistory : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            // Also migrate History records where new SourceGuid is still null guid after previous migrations.
```
jsoi-eq, line 13: "null guid" means "00000000-0000-0000-0000-000000000000"?
If so, a little confusing to me, as the value is not "null".
Maybe default guid or empty guid or zero guid or similiar would make more sense.

jsoi-eq (approved):

src/Equinor.ProCoSys.Preservation.Infrastructure/Migrations/20230607071458_MigrateNullGuidsInHistory.cs
janid1967: Agree, should have used term zero-guid. To much work to rename in code now. I rename PR title at least

PR #943, opened 2023-09-06
Delete .github/dependabot.yml

TordJoranger:


janid1967 (approved):

janid1967:
Also delete all PR's made by dependabot?

TordJoranger:
Yeah, I don't see why not. We can update NuGet packages regardless.

PR #951, opened 2023-10-25
Use GUID instead og OLD-values

terjewoldseth:
This will remove the need for old-values in Tag-messages.

McPkg messages and CommPkg messages would still need old-values.  A new feature will be added that enables the removal of this requirement too.

### DIFF
--- a/src/Equinor.ProCoSys.Preservation.WebApi/Synchronization/BusReceiverService.cs
+++ b/src/Equinor.ProCoSys.Preservation.WebApi/Synchronization/BusReceiverService.cs
@@ -120,21 +120,8 @@ namespace Equinor.ProCoSys.Preservation.WebApi.Synchronization
             TrackTagEvent(tagEvent);
 
             _plantSetter.SetPlant(tagEvent.Plant);
-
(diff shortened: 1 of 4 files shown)


src/Equinor.ProCoSys.Preservation.WebApi/Synchronization/BusReceiverService.cs
```diff
@@ -143,11 +130,10 @@ private async Task ProcessTagEventAsync(TagTopic tagEvent, string messageJson)
                     tagToUpdate.Rename(tagEvent.TagNo);
                 }
 
-                if (!tagEvent.ProjectNameOld.IsEmpty() &&
-                    tagEvent.ProjectName != tagEvent.ProjectNameOld)
+                var project = await _projectRepository.GetProjectOnlyByTagGuidAsync(tagGuid);
```
TordJoranger, line 133: we're calling SingleOrDefault, so project can be null here. I'm uncertain if it should, if not maybe we change to Single(),
otherwise we need nullcheck.

TordJoranger (approved):
with suggestions

PR #958, opened 2024-06-28
111982 - Fam transfer

haavard-hoijord:
[AB#111982](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/111982)
https://github.com/equinor/cc-toolbox/issues/357


src/Equinor.ProCoSys.Preservation.Command/ActionCommands/CreateAction/CreateActionCommandHandler.cs
```diff
@@ -26,6 +28,7 @@ public async Task<Result<int>> Handle(CreateActionCommand request, CancellationT
             var tag = await _projectRepository.GetTagWithActionsByTagIdAsync(request.TagId);
 
             var actionToAdd = new Action(
+                new Guid(),
```
jsoi-eq, line 31: Guid property ought to be set in Action constructor instead of as parameter in the constructor, see IPO Invitation for example.

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/ActionAddedEventHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        _integrationEventPublisher = integrationEventPublisher;
+        _createEventHelper = createEventHelper;
+        _context = context;
+    }
+
+    public async Task Handle(ActionAddedEvent notification, CancellationToken cancellationToken)
+    {
+        var tagObject = (from tag in _context.QuerySet<Tag>()
+                                        where tag.Guid == notification.SourceGuid
```
jsoi-eq, line 32: I think notification Source.Guid refers to the guid of the Action, and not the Tag, so it cannot be used this way.
ChatGpt says we can find TagId using the following:
     var tagId = _context.Entry(action).Property<int>("TagId").CurrentValue;
and retrieve the tab object using
    var tag = _context.Tags.Find(tagId);
Seems a tiny bit hacky to me, so would like to think it over.

src/Equinor.ProCoSys.Preservation.Command/Events/ActionEvent.cs
```diff
@@ -0,0 +1,31 @@
...
+        DueDate = dueDate;
+        Overdue = overdue;
+        Closed = closed;
+    }
+
+    public Guid Guid { get; }
+    public Guid ProCoSysGuid => Guid;
+    public string Plant { get; }
+    public string ProjectName { get; }
+    public string TagNr { get; }
```
jsoi-eq, line 25: We don't need TagNr as this will be joined in on FAM side from their Tag table.
But we do need the TagGuid instead, so they can perform that join.

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/ProjectAggregate/Action.cs
```diff
@@ -21,11 +21,19 @@ protected Action() : base(null)
...
+            ObjectGuid = Guid;
+            ProCoSysGuid = Guid;
+
             Title = title;
             Description = description;
             SetDueTime(dueTimeUtc);
         }
 
+        public Guid Guid { get; private set; }
+        public Guid ObjectGuid { get; private set; }
```
jsoi-eq, line 34: ObjectGuid and ProCoSysGuid can be removed.
I know I mentioned ProCoSysGuid, so my mistake, it only needs to exist in the message sent to ServiceBus...

src/Equinor.ProCoSys.Preservation.MessageContracts/IActionEventV1.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Equinor.ProCoSys.Preservation.MessageContracts;
+
+public interface IActionEventV1 : IIntegrationEvent
+{
+    string TagNr { get; }
```
jsoi-eq, line 5: Replace with TagGuid

src/Equinor.ProCoSys.Preservation.WebApi/Startup.cs
```diff
@@ -176,7 +176,7 @@ public void ConfigureServices(IServiceCollection services)
 
                 services.AddPcsServiceBusIntegration(options => options
                     .UseBusConnection(Configuration.GetConnectionString("ServiceBus"))
-                    .WithLeaderElector(leaderElectorUrl)
+                   // .WithLeaderElector(leaderElectorUrl)
```
jsoi-eq, line 179: I think it works even with this commented in.
At least I got it working in IPO even though it spits out exceptions with a "LeaderElectorUrl": "https://fake.host.com" config.
Just a bit nervous we forget to comment it back in again before deploy...

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+        var project = await _projectRepository.GetProjectOnlyByTagGuidAsync(tag.Guid);
+
+        return new ActionEvent(
+            action.Guid,
+            action.Plant,
+            project.Name,
+            tag.TagNo,
+            action.Title,
+            action.Description,
+            action.DueTimeUtc != null ? DateOnly.FromDateTime(action.DueTimeUtc.GetValueOrDefault(DateTime.Now)) : null,
```
jsoi-eq, line 31: Defaulting to DateTime.Now risks sending over incorrect date, however unlikely.
Better to just use action.DueTimeUtc.Value since we have already done a null check?

jsoi-eq (changes requested):
See comments.

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/ProjectAggregate/Action.cs
haavard-hoijord: Well Guid is required for the IHaveGuid interface, but i can get rid of the ObjectGuid

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/ProjectAggregate/Action.cs
```diff
@@ -22,7 +22,6 @@ public Action(string plant, string title, string description, DateTime? dueTimeU
             : base(plant)
         {
             Guid = Guid.NewGuid();
-            ObjectGuid = Guid;
             ProCoSysGuid = Guid;
```
jsoi-eq, line 25: What I meant was to keep Guid property but remove both ObjectGuid and ProCoSysGuid. 
We only need a property ProCoSysGuid when sending to FAM, and don't need it in the aggregate, that was my miscommunication last time around.

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/ActionClosedEventHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+        _createEventHelper = createEventHelper;
+        _context = context;
+        _projectRepository = projectRepository;
+    }
+
+    public async Task Handle(ActionClosedEvent notification, CancellationToken cancellationToken)
+    {
+        var tagObject = await _projectRepository.GetTagOnlyByGuidAsync(notification.SourceGuid);
+
+        var actionObject = await (from action in _context.QuerySet<Action>()
```
jsoi-eq, line 38: I realized that in IPO latest version we are sending in the whole object in the Notification.
So instead of sending in actionGuid, you can send in the entire Action object.
Then you don't need to query for it again in this method.

haavard-hoijord: I am not quite sure i understood what you meant here

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -12,20 +12,17 @@ public class CreateEventHelper : ICreateEventHelper
...
     private readonly IProjectRepository _projectRepository;
 
-    public CreateEventHelper(IProjectRepository projectRepository)
-    {
-        _projectRepository = projectRepository;
-    }
+    public CreateEventHelper(IProjectRepository projectRepository) => _projectRepository = projectRepository;
 
-    public async Task<IActionEventV1> CreateActionEvent(Action action, Tag tag)
+    public async Task<IActionEventV1> CreateActionEvent(Action action, Guid tagId)
```
jsoi-eq, line 17: Could you use tagGuid as the parameter name instead?
We have two "id" columns in Tag, Id and Guid.
For me I get confused with the name tagId since I assume this refers to Tag.Id, and not Tag.Guid.
The same tagId -> tagGuid applies for some of the other files below.

Also matches with the existing convention where sourceGuid is used instead of sourceId

jsoi-eq (changes requested):
Looks good, only the tagId -> tagGuid rename, left.

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+            tagGuid,
+            action.Title,
+            action.Description,
+            action.DueTimeUtc != null ? DateOnly.FromDateTime(action.DueTimeUtc.Value) : null,
+            action.IsOverDue(),
+            action.ClosedAtUtc != null ? DateOnly.FromDateTime(action.ClosedAtUtc.Value) : null
+            );
+    }
+
+    public async Task SendTagRequirementEvents(TagRequirement tagRequirement, Guid tagGuid, CancellationToken cancellationToken)
```
jsoi-eq, line 44: This method is not in use and can be removed?

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+            Plant = tagRequirement.Plant,
+            ProjectName = project.Name,
+            IntervalWeeks = tagRequirement.IntervalWeeks,
+            Usage = tagRequirement.Usage.ToString(),
+            NextDueTimeUtc = tagRequirement.NextDueTimeUtc,
+            IsVoided = tagRequirement.IsVoided,
+            IsInUse = tagRequirement.IsInUse,
+            RequirementDefinitionGuid = tagRequirement.RequirementDefinitionGuid,
+            CreatedAtUtc = tagRequirement.CreatedAtUtc,
+            CreatedById = tagRequirement.CreatedById,
```
jsoi-eq, line 66: Instead of TagRequirement.CreatedById, change this to TagRequirement.CreatedByGuid instead. 
So you need to read Guid property  value from the Person table based on the CreatedById, see IPO code for an example.
This guid contains the Azure Oid of a user, and will be used in FAM for a lookup on their side to get name and email-address etc.

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+            ModifiedAtUtc = tagRequirement.ModifiedAtUtc,
+            ModifiedById = tagRequirement.ModifiedById,
```
jsoi-eq, line 68: => ModifiedByGuid

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+            ReadyToBePreserved = tagRequirement.ReadyToBePreserved
+        };
+    }
+
+    public async Task<IPreservationPeriodEventV1> CreatePreservationPeriodEvent(PreservationPeriod preservationPeriod)
```
jsoi-eq, line 73: Remove async if we don't use any await.

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+            Comment = preservationPeriod.Comment,
+            PreservationRecordGuid = preservationPeriod.PreservationRecord.Guid,
+            CreatedAtUtc = preservationPeriod.CreatedAtUtc,
+            CreatedById = preservationPeriod.CreatedById,
+            ModifiedAtUtc = preservationPeriod.ModifiedAtUtc,
+            ModifiedById = preservationPeriod.ModifiedById,
+            TagRequirementGuid = preservationPeriod.TagRequirementGuid,
+            Guid = preservationPeriod.Guid,
+            Plant = preservationPeriod.Plant,
+            PreservationRecord =
```
jsoi-eq, line 89: Don't use nested structures, we don't handle those in FamWebJob.
Is PreservedById used for a join somewhere else? We usually don't send over internal int id values unless there is a specific reason, like in IPO we send it over since it is used for generating a title.
Curious: What was the motivation for doing these as a dynamic?

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+                BulkPreserved = preservationRecord.BulkPreserved
+            }
+        };
+    }
+
+    public async Task<IFieldEventV1> CreateFieldEvent(Field field, Guid sourceGuid) =>
+        new FieldEvent
+        {
+            RequirementDefinitionGuid = sourceGuid,
+            FieldId = field.Id,
```
jsoi-eq, line 102: I don't think we need FieldId since we have Guid here?

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+            PreservationRecord =
+            {
+                PreservedAtUtc = preservationRecord.PreservedAtUtc,
+                PreservedById = preservationRecord.PreservedById,
+                BulkPreserved = preservationRecord.BulkPreserved
+            }
+        };
+    }
+
+    public async Task<IFieldEventV1> CreateFieldEvent(Field field, Guid sourceGuid) =>
```
jsoi-eq, line 98: Remove async

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+        new FieldEvent
+        {
+            RequirementDefinitionGuid = sourceGuid,
+            FieldId = field.Id,
+            Label = field.Label,
+            Unit = field.Unit,
+            SortKey = field.SortKey,
+            FieldType = field.FieldType.ToString(),
+            CreatedAtUtc = field.CreatedAtUtc,
+            CreatedById = field.CreatedById,
```
jsoi-eq, line 108: => CreatedByGuid

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+            ModifiedAtUtc = field.ModifiedAtUtc,
+            ModifiedById = field.ModifiedById,
```
jsoi-eq, line 110: ModifiedByGuid

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+            Guid = field.Guid,
+            Plant = field.Plant
+        };
+
+    public async Task<IRequirementDefinitionEventV1> CreateRequirementDefinitionEvent(RequirementDefinition requirementDefinition) =>
```
jsoi-eq, line 115: Remove async

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+            Usage = requirementDefinition.Usage.ToString(),
+            SortKey = requirementDefinition.SortKey,
+            NeedsUserInput = requirementDefinition.NeedsUserInput,
+            CreatedAtUtc = requirementDefinition.CreatedAtUtc,
+            CreatedById = requirementDefinition.CreatedById,
+            ModifiedAtUtc = requirementDefinition.ModifiedAtUtc,
+            ModifiedById = requirementDefinition.ModifiedById
+        };
+
+    public async Task<IRequirementTypeEventV1> CreateRequirementTypeEvent(RequirementType requirementType) =>
```
jsoi-eq, line 132: Remove async

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+        new RequirementTypeEvent
+        {
+            Guid = requirementType.Guid,
+            Plant = requirementType.Plant,
+            Code = requirementType.Code,
+            Title = requirementType.Title,
+            IsVoided = requirementType.IsVoided,
+            SortKey = requirementType.SortKey,
+            CreatedAtUtc = requirementType.CreatedAtUtc,
+            CreatedById = requirementType.CreatedById,
```
jsoi-eq, line 142: CreatedByGuid

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/CreateEventHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+            Title = requirementDefinition.Title,
+            IsVoided = requirementDefinition.IsVoided,
+            DefaultIntervalWeeks = requirementDefinition.DefaultIntervalWeeks,
+            Usage = requirementDefinition.Usage.ToString(),
+            SortKey = requirementDefinition.SortKey,
+            NeedsUserInput = requirementDefinition.NeedsUserInput,
+            CreatedAtUtc = requirementDefinition.CreatedAtUtc,
+            CreatedById = requirementDefinition.CreatedById,
+            ModifiedAtUtc = requirementDefinition.ModifiedAtUtc,
+            ModifiedById = requirementDefinition.ModifiedById
```
jsoi-eq, line 129: ModifiedByGuid

src/Equinor.ProCoSys.Preservation.Command/Events/DeleteEvent.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using Equinor.ProCoSys.Preservation.MessageContracts;
+
+namespace Equinor.ProCoSys.Preservation.Command.Events;
+
+public class DeleteEvent : IDeleteEventV1
```
jsoi-eq, line 6: Is this published to the correct topic on servicebus?
An ActionDelete message should be published to the action topic, TagDelete to Tag topic etc, will this happen when using a generic DeleteEvent like this?

src/Equinor.ProCoSys.Preservation.Command/Events/FieldEvent.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using Equinor.ProCoSys.Preservation.MessageContracts;
+
+namespace Equinor.ProCoSys.Preservation.Command.Events;
+
+public class FieldEvent : IFieldEventV1
+{
+    public Guid RequirementDefinitionGuid { get; set; }
+    public int FieldId { get; set; }
```
jsoi-eq, line 9: Not required since we have a Guid property?

src/Equinor.ProCoSys.Preservation.Command/Events/FieldEvent.cs
```diff
@@ -0,0 +1,24 @@
...
+    public string Label { get; set; }
+    public string Unit { get; set; }
+    public int SortKey { get; set; }
+    public bool? ShowPrevious { get; set; }
+    public string FieldType { get; set; }
+    public DateTime CreatedAtUtc { get; set; }
+    public int CreatedById { get; set; }
```
jsoi-eq, line 16: You know the drill by now ;), I will not write any more comments for CreatedById and ModifiedById from now onwards.

src/Equinor.ProCoSys.Preservation.Command/Events/PreservationPeriodsEvent.cs
```diff
@@ -0,0 +1,23 @@
...
+    public string Comment { get; set; }
+    public Guid PreservationRecordGuid { get; set; }
+    public DateTime CreatedAtUtc { get; set; }
+    public int CreatedById { get; set; }
+    public DateTime? ModifiedAtUtc { get; set; }
+    public int? ModifiedById { get; set; }
+    public Guid Guid { get; set; }
+    public Guid ProCoSysGuid => Guid;
+    public string Plant { get; set; }
+    public string ProjectName { get; set; } = "";
```
jsoi-eq, line 21: If we don't have a projectname, I think we can remove the property unless there is a reason to keep it?

src/Equinor.ProCoSys.Preservation.Command/Events/PreservationRecordEvent.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using Equinor.ProCoSys.Preservation.MessageContracts;
+
+namespace Equinor.ProCoSys.Preservation.Command.Events;
+
+public class PreservationRecordEvent : IPreservationRecordEventV1
```
jsoi-eq, line 6: Need a Guid property. 
May need a Plant property, not sure, we need to look into that, as I realize this is across all plants master data.

src/Equinor.ProCoSys.Preservation.Command/Events/PreservationRecordEvent.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    public DateTime PreservedAtUtc { get; set; }
+    public int PreservedById { get; set; }
```
jsoi-eq, line 9: PreservedByGuid, in case you did a search replace for CreatedById and ModifiedById and missed this one.

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/ProjectAggregate/Action.cs
```diff
@@ -21,11 +22,16 @@ protected Action() : base(null)
...
         {
+            Guid = Guid.NewGuid();
+
             Title = title;
             Description = description;
             SetDueTime(dueTimeUtc);
         }
 
+        public Guid Guid { get; private set; }
+        public Guid TagGuid { get; set; }
```
jsoi-eq, line 33: I need to think a little bit about this one. 
We already have a reference to Tag with TagId, although I see it is hidden.
Introducing this would lead to an extra column in the database, and us having basically two references to Tag, one by Id and one by Guid.
If we do add this, at least it has to be non-nullable, which also leads to a job populating this column in database with correct values, and having a constraint so we can't add values not existing in Tag.Guid column.

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/ProjectAggregate/PreservationPeriod.cs
```diff
@@ -42,7 +44,11 @@ public PreservationPeriod(string plant, int intervalWeeks, PreservationPeriodSta
         public int CreatedById { get; private set; }
         public DateTime? ModifiedAtUtc { get; private set; }
         public int? ModifiedById { get; private set; }
+
         public int TagRequirementId { get; set; }
+        public Guid TagRequirementGuid { get; set; }
```
jsoi-eq, line 49: Reminder to self to think about about whether we want the extra TagRequirementGuid column.

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/ProjectAggregate/TagRequirement.cs
```diff
@@ -41,18 +43,24 @@ public TagRequirement(string plant, int intervalWeeks, RequirementDefinition req
...
+        public Guid Guid { get; private set; }
+        public Guid TagGuid { get; set; }
+
         public int IntervalWeeks { get; private set; }
         public RequirementUsage Usage { get; private set; }
         public DateTime? NextDueTimeUtc { get; private set; }
         public bool IsVoided { get; set; }
         public bool IsInUse => _preservationPeriods.Any();
         public int RequirementDefinitionId { get; private set; }
+        public Guid RequirementDefinitionGuid { get; private set; }
```
jsoi-eq, line 62: Note to self: Consider whether we want this property.

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/RequirementTypeAggregate/Field.cs
```diff
@@ -50,6 +53,8 @@ protected Field() : base(null)
         public int CreatedById { get; private set; }
         public DateTime? ModifiedAtUtc { get; private set; }
         public int? ModifiedById { get; private set; }
+        public Guid Guid { get; private set; }
+        public Guid RequirementDefinitionGuid { get; set; }
```
jsoi-eq, line 57: Note to self: Consider whether we want this property.

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/RequirementTypeAggregate/RequirementDefinition.cs
```diff
@@ -23,12 +24,17 @@ protected RequirementDefinition()
...
+            Guid = Guid.NewGuid();
+
             Title = title;
             DefaultIntervalWeeks = defaultIntervalWeeks;
             Usage = usage;
             SortKey = sortKey;
         }
 
+        public Guid Guid { get; private set; }
+        public Guid RequirementTypeGuid { get; set; }
```
jsoi-eq, line 36: Note to self: vConsider whether we want this property.

src/Equinor.ProCoSys.Preservation.Domain/Events/TagRequirementAddedEvent.cs
```diff
@@ -1,19 +1,21 @@
...
             SourceGuid = sourceGuid;
-            RequirementDefinitionId = requirementDefinitionId;
+            TagRequirement = tagRequirement;
         }
 
         public string Plant { get; }
         public Guid SourceGuid { get; }
-        public int RequirementDefinitionId { get; }
+        public TagRequirement TagRequirement { get; }
+        public int RequirementDefinitionId => TagRequirement.RequirementDefinitionId;
```
jsoi-eq, line 19: Was there a reason for Id instead of Guid?

src/Equinor.ProCoSys.Preservation.Domain/Events/TagRequirementDeletedEvent.cs
```diff
@@ -1,19 +1,21 @@
...
             SourceGuid = sourceGuid;
-            RequirementDefinitionId = requirementDefinitionId;
+            TagRequirement = tagRequirement;
         }
 
         public string Plant { get; }
         public Guid SourceGuid { get; }
-        public int RequirementDefinitionId { get; }
+        public TagRequirement TagRequirement { get; }
+        public int RequirementDefinitionId => TagRequirement.RequirementDefinitionId;
```
jsoi-eq, line 19: Was there a reason for Id instead of Guid?

src/Equinor.ProCoSys.Preservation.Domain/Events/TagRequirementPreservedEvent.cs
```diff
@@ -8,19 +9,20 @@ public class TagRequirementPreservedEvent : IDomainEvent
...
-            RequirementDefinitionId = requirementDefinitionId;
+            TagRequirement = tagRequirement;
             DueInWeeks = dueInWeeks;
             PreservationRecordGuid = preservationRecordGuid;
         }
         public string Plant { get; }
         public Guid SourceGuid { get; }
-        public int RequirementDefinitionId { get; }
+        public TagRequirement TagRequirement { get; }
+        public int RequirementDefinitionId => TagRequirement.RequirementDefinitionId;
```
jsoi-eq, line 25: Was there a reason for Id instead of Guid?

src/Equinor.ProCoSys.Preservation.Domain/Events/TagRequirementUnvoidedEvent.cs
```diff
@@ -1,19 +1,21 @@
...
             SourceGuid = sourceGuid;
-            RequirementDefinitionId = requirementDefinitionId;
+            TagRequirement = tagRequirement;
         }
 
         public string Plant { get; }
         public Guid SourceGuid { get; }
-        public int RequirementDefinitionId { get; }
+        public TagRequirement TagRequirement { get; }
+        public int RequirementDefinitionId => TagRequirement.RequirementDefinitionId;
```
jsoi-eq, line 19: Was there a reason for Id instead of Guid?

src/Equinor.ProCoSys.Preservation.Domain/Events/TagRequirementUpdatedEvent.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+            Plant = plant;
+            SourceGuid = sourceGuid;
+            TagRequirement = tagRequirement;
+        }
+
+        public string Plant { get; }
+        public Guid SourceGuid { get; }
+        public TagRequirement TagRequirement { get; }
+        public int RequirementDefinitionId => TagRequirement.RequirementDefinitionId;
```
jsoi-eq, line 19: Was there a reason for Id instead of Guid?

src/Equinor.ProCoSys.Preservation.Infrastructure/Migrations/20240701141306_AddMassTransitAndGuids.cs
```diff
@@ -0,0 +1,155 @@
...
+
+namespace Equinor.ProCoSys.Preservation.Infrastructure.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddMassTransitAndGuids : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<Guid>(
```
jsoi-eq, line 14: Note to self: Consider removing the new Guid columns from here and let MassTransit be one migration and Guids a different migration.

src/Equinor.ProCoSys.Preservation.Infrastructure/Migrations/20240702092142_AddGuidsToRequirementsAndField.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Equinor.ProCoSys.Preservation.Infrastructure.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddGuidsToRequirementsAndField : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<Guid>(
```
jsoi-eq, line 14: Note to self: All these Guids need to be populated with initial values

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/ProjectAggregate/Tag.cs
```diff
@@ -246,13 +246,14 @@ public void AddRequirement(TagRequirement tagRequirement)
...
+            tagRequirement.TagGuid = Guid;
+
             _requirements.Add(tagRequirement);
             if (Status == PreservationStatus.Active || Status == PreservationStatus.InService)
             {
                 tagRequirement.StartPreservation();
             }
             UpdateNextDueTimeUtc();
-            AddDomainEvent(new TagRequirementAddedEvent(tagRequirement.Plant, Guid, tagRequirement.RequirementDefinitionId));
         }
```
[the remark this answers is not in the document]
haavard-hoijord: Yes, so this change here just moved the added event from the AddRequirement to the SetCreated method, which is activated by changes in the database instead of the AddRequirement function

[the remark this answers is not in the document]
jsoi-eq: Regarding changes for Area or Discipline, I see your point for SetModified which sends out a TagUpdatedEvent when a change has happened, and for new ones we have TagCreatedEvent in constructor, and I agree.

And I see now what you did with TagRequirementAddedEvent, and I agree it makes sense to move it, although I wonder if it makes more sense to put it in the constructor in TagRequirement, similar to what is done with TagCreatedEvent in Tag.

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/HistoryEvents/ActionAddedEventHandler.cs
```diff
@@ -1,25 +1,39 @@
...
 namespace Equinor.ProCoSys.Preservation.Command.EventHandlers.HistoryEvents
 {
     public class ActionAddedEventHandler : INotificationHandler<ActionAddedEvent>
     {
         private readonly IHistoryRepository _historyRepository;
+        private readonly IProjectRepository _projectRepository;
+        private readonly PreservationContext _context;
 
-        public ActionAddedEventHandler(IHistoryRepository historyRepository) => _historyRepository = historyRepository;
+        public ActionAddedEventHandler(IHistoryRepository historyRepository, IProjectRepository projectRepository, PreservationContext context)
```
jsoi-eq, line 20: It seems from other places in the code that we should not use PreseverationContext directly, but rather IReadOnlyContext. Then we also don't need the dependency to Infrastructure which was added.
So something like 
            var tagId = await _readOnlyContext.QuerySet<Action>().Where(s => s.Guid == notification.TagRequirement.Guid)
                .Select(s => EF.Property<int>(s, "TagId")).SingleAsync(cancellationToken: cancellationToken);

Hopefully works?
Applies for all HistoryEvents that you changed.

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/ProjectAggregate/Tag.cs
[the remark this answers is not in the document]
haavard-hoijord: Well putting it in the constructor does have a theoretical issue in that the event could be run before it has actually been saved to the database so querying for the tag id or similar would fail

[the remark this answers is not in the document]
jsoi-eq: Hmm, so you're saying that if we are creating a new tag, and then adding a requirement, the tag will have Id = 0 because a savechanges has not been called to populate the Id value, sounds plausible, good catch.
And in SetCreated we know savechanges has been called so we do have an Id at that point in time.
I retract my suggestion to move it to constructor.
Although maybe add a comment since this was non-intuitive?

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/PreservationPeriodDeletedEventHandler.cs
```diff
@@ -15,7 +15,7 @@ public class PreservationPeriodDeletedEventHandler  : INotificationHandler<Prese
 
     public async Task Handle(PreservationPeriodDeletedEvent notification, CancellationToken cancellationToken)
     {
-        var deleteEvent = new DeleteEvent(notification.PreservationPeriod.Guid, notification.PreservationPeriod.Plant, "");
+        var deleteEvent = new DeleteEvent.PreservationPeriodDeleteEvent(notification.PreservationPeriod.Guid, notification.PreservationPeriod.Plant, "");
```
jsoi-eq, line 18: I think NULL may be a better representation than empty string, when it comes to the FAM database.

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/RequirementDefinitionDeletedEventHandler.cs
```diff
@@ -15,7 +15,7 @@ public class RequirementDefinitionDeletedEventHandler  : INotificationHandler<Re
 
     public async Task Handle(RequirementDefinitionDeletedEvent notification, CancellationToken cancellationToken)
     {
-        var deleteEvent = new DeleteEvent(notification.RequirementDefinition.Guid, notification.RequirementDefinition.Plant, "");
+        var deleteEvent = new DeleteEvent.RequirementDefinitionDeleteEvent(notification.RequirementDefinition.Guid, notification.RequirementDefinition.Plant, "");
```
jsoi-eq, line 18: Same

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/RequirementFieldDeletedEventHandler.cs
```diff
@@ -15,7 +15,7 @@ public class RequirementFieldDeletedEventHandler  : INotificationHandler<Require
 
     public async Task Handle(RequirementDeletedFieldEvent notification, CancellationToken cancellationToken)
     {
-        var deleteEvent = new DeleteEvent(notification.Field.Guid, notification.Field.Plant, "");
+        var deleteEvent = new DeleteEvent.RequirementFieldDeleteEvent(notification.Field.Guid, notification.Field.Plant, "");
```
jsoi-eq, line 18: Same

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/RequirementTypeDeletedEventHandler.cs
```diff
@@ -15,7 +15,7 @@ public class RequirementTypeDeletedEventHandler  : INotificationHandler<Requirem
 
     public async Task Handle(RequirementTypeDeletedEvent notification, CancellationToken cancellationToken)
     {
-        var deleteEvent = new DeleteEvent(notification.RequirementType.Guid, notification.RequirementType.Plant, "");
+        var deleteEvent = new DeleteEvent.RequirementTypeDeleteEvent(notification.RequirementType.Guid, notification.RequirementType.Plant, "");
```
jsoi-eq, line 18: Same

src/Equinor.ProCoSys.Preservation.Command/EventPublishers/BusEventMessage.cs
```diff
@@ -5,9 +5,9 @@ namespace Equinor.ProCoSys.Preservation.Command.EventPublishers;
 
 public class BusEventMessage : IIntegrationEvent
 {
-    public Guid Guid { get; set;}
+    public Guid Guid { get; init;}
     public Guid ProCoSysGuid => Guid;
```
jsoi-eq, line 9: I see IIntegrationEvent received 
Guid ProCoSysGuid => Guid;
Which is fine.
But do we need this line anymore in BusEventMessage then?

src/Equinor.ProCoSys.Preservation.Command/Events/DeleteEvent.cs
richard1927: It does not look like IDeleteEventV1 events are handled, so they will not be published to any topic. I think they would throw an error.

richard1927: Actually, it does not look like they would throw an error. It looks like all events go to the "IIntegrationEvent" topic. I will investigate further.

richard1927: It looks like the EntityNames attributes specify what queues the messages go to. ActionDelete should go to the PreservationAction topic, and TagDelete should go to the PreservationTag topic.

richard1927: Local testing results in a argument exception when evaluating what topic to use. I will fix this.

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/ProjectAggregate/Tag.cs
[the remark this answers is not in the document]
richard1927: Resolving conversation. This will be refactored completely when making the change to pre save events.

jsoi-eq (approved):

PR #959, opened 2024-09-27
Fam transfer refactor

richard1927:
- Removing the need for semaphore threading locks by handling events synchronously
- Removing duplicate boiler plate code by re-using logic
- Explicitly setting all event service bus destinations via EntityName attributes
- Fixing failing unit tests


src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateFieldEventHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+        var createdBy = await _personRepository.GetReadOnlyByIdAsync(entity.CreatedById);
+        var modifiedBy = entity.ModifiedById.HasValue ? await _personRepository.GetReadOnlyByIdAsync(entity.ModifiedById.Value) : null;
+
+        return new FieldEvent
+        {
+            RequirementDefinitionGuid = requirementDefinition.Guid,
+            Label = entity.Label,
+            Unit = entity.Unit,
+            SortKey = entity.SortKey,
+            FieldType = entity.FieldType.ToString(),
```
jsoi-eq, line 31: I haven't thought too much about consequences and solutions, but if someone decides to change the name of an enum field, i.e. FieldType.Info to FieldType.Information this will cause issues.
I guess there are tricks for assigning an enum field a string value through attributes or something, or do we just accept the risk, what do you think?

richard1927: Good catch. I just refactored what was already written. I agree, this can be done better. I will look into this.

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateTagEventHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public async Task<TagEvent> CreateEvent(Tag entity)
+    {
+        var project = await _projectRepository.GetProjectOnlyByTagGuidAsync(entity.Guid);
+        var createdBy = await _personRepository.GetReadOnlyByIdAsync(entity.CreatedById);
+        var modifiedBy = entity.ModifiedById.HasValue ? await _personRepository.GetReadOnlyByIdAsync(entity.ModifiedById.Value) : null;
+        var step = await _journeyRepository.GetStepByStepIdAsync(entity.StepId);
+
+        return new TagEvent
```
jsoi-eq, line 32: We need to think a little bit about which fields to return here.
Tag data is transferred to Preservation from Main application, so a number of fields are duplicated in Preservation and Main, following microservice thinking that duplicate data may be okay.
However, they don't want duplicated data in FAM.
One of the guidelines we have received is that we should only transfer data for which the application is master.
In this case, fields like AreaDescription is not mastered in Preservation, while StepGuid is.
We can discuss this.

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateResponsibleEventHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+    private readonly IPersonRepository _personRepository;
+
+    public CreateResponsibleEventHelper(IPersonRepository personRepository) => _personRepository = personRepository;
+
+    public async Task<ResponsibleEvent> CreateEvent(Responsible entity)
+    {
+        var createdBy = await _personRepository.GetReadOnlyByIdAsync(entity.CreatedById);
+        var modifiedBy = entity.ModifiedById.HasValue ? await _personRepository.GetReadOnlyByIdAsync(entity.ModifiedById.Value) : null;
+
+        return new ResponsibleEvent
```
jsoi-eq, line 19: Investigate whether we receive data here from Main, and if so which data is mastered in Main and which data is mastered in Preservation.

src/Equinor.ProCoSys.Preservation.Command/Events/EntityNames/EntityNameJourneyAttribute.cs
```diff
@@ -0,0 +1,10 @@
+﻿using MassTransit;
+
+namespace Equinor.ProCoSys.Preservation.Command.Events.EntityNames;
+
+public class EntityNameJourneyAttribute : EntityNameAttribute
+{
+    public EntityNameJourneyAttribute() : base("PreservationJourneys")
```
jsoi-eq, line 7: There is a mix of singular and plurals in these attribute classes, is this intentional or not?
I.e. PreservationJourneys vs PreservationField
(I'm not familiar with the datamodel)

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateTagEventHelper.cs
richard1927: Yes, we should discuss this offline. I have questions.

src/Equinor.ProCoSys.Preservation.Command/Events/EntityNames/EntityNameJourneyAttribute.cs
richard1927: Good catch. I will unify the naming.

jsoi-eq:
Finished code review, awaiting feedback.

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateTagEventHelper.cs
richard1927: As per our discussion, I have added a task to address this concern.
https://github.com/equinor/cc-toolbox/issues/499

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateResponsibleEventHelper.cs
richard1927: This will be resolved in this task: https://github.com/equinor/cc-toolbox/issues/499

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateFieldEventHelper.cs
richard1927: After giving this some consideration, I have decided not to alter this logic too much. I think the authority of the content of the enums is outside of the responsibility of this change. If someone were to decided that the correct name for "Info" is "Information", I do not think it is the responsibility of this module to overwrite that to be "Info" instead. 

I did however change the logic to use the nameof operator instead of ToString, since it is generally reccomended practice for Enum to string conversion.

src/Equinor.ProCoSys.Preservation.Command/Events/EntityNames/EntityNameJourneyAttribute.cs
richard1927: All attributes are now singular

jsoi-eq (approved):

PR #969, opened 2024-10-28
648 mode pre save events

richard1927:
Refactoring mode integration events to be send prior to changes being saved to the database. This is done to follow the outbox pattern.

https://github.com/equinor/cc-toolbox/issues/648


jsoi-eq (approved):

src/Equinor.ProCoSys.Preservation.Domain/AggregateModels/JourneyAggregate/Step.cs
```diff
@@ -85,7 +84,7 @@ public void SetCreated(Person createdBy)
 
             CreatedById = createdBy.Id;
 
-            AddPostSaveDomainEvent(new StepPostSaveEvent(this));
+            AddDomainEvent(new PlantEntityCreatedEvent<Step>(this));
```
TordJoranger, line 87: @richard1927 Can we discuss this change? 
To me  PlantEntityCreatedEvent<Step> would indicate that a Step has been created, and I would assume it was added in the constructor or similar. This seems to be CreatedByPerson being set on the entity?

PR #974, opened 2024-10-31
Removing unowned data from integration events

richard1927:
Data is received by Preservation over service bus and is used internally. Some of that data was included in integration events that are sent to Alpha. Since Preservation is not the source/owner of that information, that data should not be included in the events.


src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateResponsibleEventHelper.cs
```diff
@@ -20,8 +20,6 @@ public async Task<ResponsibleEvent> CreateEvent(Responsible entity)
         {
             ProCoSysGuid = entity.Guid,
             Plant = entity.Plant,
-            Code = entity.Code,
```
jsoi-eq, line 23: I'm thinking maybe we have to include Code in the entities.
Reason being if they want to retrieve the code, we don't have enough info in the object to allow them to make a join on their side.
But we can discuss this with Peter Øvstedal.

jsoi-eq (approved):

richard1927:
https://github.com/equinor/cc-toolbox/issues/499

PR #980, opened 2024-11-21
Updating test packages

richard1927:
Running testas was being problematic. Updating the packages resolved the issue.

This is most likely related to recent .net 8 updates.

https://github.com/equinor/cc-toolbox/issues/792


jsoi-eq (commented):
We need to be careful with Moq upgrades, there was one version where it hoovered up emailaddresses on local machine and sent to the developer.
https://medium.com/@atakankorez/what-happened-in-moq-library-a-blog-on-controversies-832629a125ef

richard1927:
> We need to be careful with Moq upgrades, there was one version where it hoovered up emailaddresses on local machine and sent to the developer. https://medium.com/@atakankorez/what-happened-in-moq-library-a-blog-on-controversies-832629a125ef

Interesting case, but I do not know how we should address this concern. If mock is so unreliable that we cannot safely update a minor version of the package, we should remove/replace it.

I will discuss this with you on Monday.

jsoi-eq (approved):

PR #981, opened 2024-11-25
799 step event journey ID

richard1927:
Adding journey ID to step events.

Since steps do not contain any journey information, the event logic needed to be refactored over to journey child events.


src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateStepEventHelper.cs
```diff
@@ -1,48 +1,19 @@
...
-            SortKey = entity.SortKey,
-            IsVoided = entity.IsVoided,
-            CreatedAtUtc = entity.CreatedAtUtc,
-            CreatedByGuid = createdBy.Guid,
-            ModifiedAtUtc = entity.ModifiedAtUtc,
-            ModifiedByGuid = modifiedBy?.Guid
-        };
+        var journeys = await journeyRepository.GetJourneysByStepIdsAsync([entity.Id]);
+        
+        return await createActionEventHelper.CreateEvent(journeys.First(), entity);
```
jsoi-eq, line 17: I assume journeys will never be null.
But is there a reason to use First() instead of Single()?

jsoi-eq (approved):

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateStepEventHelper.cs
richard1927: Yes,  a step will always belong to a Journey.

First is just the operation I usually use. I see no need to ensure here that no other journey contains the step. It should not happen, but i think this is not the place to enforce that.

PR #986, opened 2024-12-02
824 field deletion events

richard1927:
Emitting deletion events for fields when its parent requirement definition is deleted

https://github.com/equinor/cc-toolbox/issues/824


src/tests/Equinor.ProCoSys.Preservation.Command.Tests/EventHandlers/IntegrationEvents/RequirementTypeDeletedEventHandlerTests.cs
```diff
@@ -18,15 +14,17 @@ public class RequirementTypeDeletedEventHandlerTests
...
     private RequirementTypeDeletedEventHandler _dut;
-    private List<IIntegrationEvent> _publishedEvents = new();
+    private bool _eventPublished;
 
     [TestInitialize]
     public void Setup()
     {
         // Arrange
         var mockPublisher = new Mock<IIntegrationEventPublisher>();
         mockPublisher.Setup(x => x.PublishAsync(It.IsAny<IIntegrationEvent>(), default))
```
jsoi-eq, line 24: I'm wondering whether it would be better to use a specific type, e.g. RequirementTypeDeleteEvent instead of IIntegrationEvent to ensure the correct type of event is published, what do you think?

Although I'm not able to make it work myself when I change the code to that...

src/tests/Equinor.ProCoSys.Preservation.Command.Tests/EventHandlers/IntegrationEvents/CreateRequirementTypeDeletedEventHelperTests.cs
```diff
@@ -18,36 +17,55 @@ public void Setup()
...
         
         // Act
-        var integrationEvent = CreateRequirementTypeDeletedEventHelper.CreateEvent(_requirementType);
-        var result = integrationEvent.ProCoSysGuid;
+        var integrationEvents = CreateRequirementTypeDeletedEventHelper.CreateEvents(_requirementType);
+        var deletionEvent = integrationEvents.Single(e => e.GetType() == typeof(RequirementTypeDeleteEvent));
+        var result = deletionEvent.ProCoSysGuid;
 
         // Assert
         Assert.AreEqual(result, expected);
```
jsoi-eq, line 58: expected and result should be in reversed order according to documentation of AreEqual

src/tests/Equinor.ProCoSys.Preservation.Command.Tests/EventHandlers/IntegrationEvents/CreateRequirementDefinitionDeletedEventHelperTests.cs
```diff
@@ -11,43 +10,61 @@ namespace Equinor.ProCoSys.Preservation.Command.Tests.EventHandlers.IntegrationE
...
         
         // Act
-        var integrationEvent = CreateRequirementDefinitionDeletedEventHelper.CreateEvent(_requirementType);
-        var result = integrationEvent.ProCoSysGuid;
+        var integrationEvents = CreateRequirementDefinitionDeletedEventHelper.CreateEvents(_requirementDefinition);
+        var deletionEvent = integrationEvents.Single(e => e.GetType() == typeof(RequirementDefinitionDeleteEvent));
+        var result = deletionEvent.ProCoSysGuid;
 
         // Assert
         Assert.AreEqual(result, expected);
```
jsoi-eq, line 57: Reverse order

src/tests/Equinor.ProCoSys.Preservation.Command.Tests/EventHandlers/IntegrationEvents/RequirementTypeDeletedEventHandlerTests.cs
richard1927: You would like to prevent publication of the interface that represent the object, and instead make a publisher for each of the integration event objects/classes?

Or are you referring to making a generic version of the publisher that can be instantiated for each of the integration events?

PR #987, opened 2024-12-03
824 Field events fixes

richard1927:
* Fixing an issue where updating fields would cause an error

The "GetRequirementDefinitionByFieldGuidAsync" in the RequirementTypeRepository was not working as intended, and only returned null values.
Fixed the issue by bypassing the need to retrieve the parent object and instead making use of the child modified event logic flow.

* Fixing an issue where sending deletion events would silently fail

Events were being sent to the publisher as a interface, IDeleteEventV1. Mass transit was unable to handle these events, and they silently were ignored.
Changing the implementation to pass the objects directly, instead of a interface representation.

https://github.com/equinor/cc-toolbox/issues/824


src/tests/Equinor.ProCoSys.Preservation.Command.Tests/EventHandlers/IntegrationEvents/RequirementTypeDeletedEventHandlerTests.cs
```diff
@@ -42,4 +42,43 @@ public async Task Handle_ShouldSendRequirementTypeDeleteEvent()
...
+        var field = new Field(TestPlant, "Test Label", FieldType.Info, 0);
+        requirementDefinition.AddField(field);
+        
+        var domainEvent = new DeletedEvent<RequirementType>(requirementType);
+
+        // Act
+        await _dut.Handle(domainEvent, default);
+
+        // Assert
+        Assert.IsTrue(_eventPublished);
```
jsoi-eq: In this test, how do we know that _eventPublished is true because a FieldDeleteEvent was published rather than a RequirementDefinitionDeleteEvent?

richard1927: Yes, that was a logic error on my part. I have updated the change with better validation.

jsoi-eq (approved):

PR #990, opened 2024-12-05
834 action deletion events

richard1927:
The solution guards against deletion of tags with actions, and therefore action deletion events "should" not be needed.
For the sake of paranoia, if somehow a tag that contains actions is deleted a corresponding action deletion event is sent.

https://github.com/equinor/cc-toolbox/issues/834


src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateTagDeleteEventHelper.cs
```diff
@@ -1,18 +1,19 @@
...
-    private readonly IProjectRepository _projectRepository;
-
-    public CreateTagDeleteEventHelper(IProjectRepository projectRepository) => _projectRepository = projectRepository;
-
-    public async Task<TagDeleteEvent> CreateEvent(Tag entity)
+    public async Task<TagDeleteEvents> CreateEvents(Tag entity)
     {
-        var project = await _projectRepository.GetProjectOnlyByTagGuidAsync(entity.Guid);
-        return new TagDeleteEvent(entity.Guid, entity.Plant, project.Name);
+        var project = await projectRepository.GetProjectOnlyByTagGuidAsync(entity.Guid);
```
jsoi-eq, line 12: Should we mix and match primary constructors / old-style constructors or stick to one style through the solution?

src/tests/Equinor.ProCoSys.Preservation.Command.Tests/EventHandlers/IntegrationEvents/TagDeletedEventHandlerTests.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        // Assert
+        CollectionAssert.Contains(types, typeof(TagDeleteEvent));
+    }
+    
+    [TestMethod]
+    public async Task Handle_ShouldSendActionDeleteEvent()
+    {
+        // Arrange
+        var tag = new Mock<Tag>();
```
jsoi-eq, line 63: Curious: Using a mock for tag, is this to save time as opposed to creating a concrete instance and having to input a lot of parameters in the constructor?

jsoi-eq (approved):

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateTagDeleteEventHelper.cs
richard1927: We should not use the primary style constructors unless we change all constructors in the solution?

jsoi-eq:
Btw, there seems to be test failures, so this would need to be fixed I guess before merging.

src/tests/Equinor.ProCoSys.Preservation.Command.Tests/EventHandlers/IntegrationEvents/TagDeletedEventHandlerTests.cs
richard1927: Yes, constructing a tag requires other objects that require other objects. In this case the content of the tag is not important, so for the sake of simple code I just mocked the Tag.

src/tests/Equinor.ProCoSys.Preservation.Command.Tests/EventHandlers/IntegrationEvents/CreateFieldDeletedEventHelperTests.cs
```diff
@@ -20,10 +20,10 @@ public void Setup()
...
     [DataTestMethod]
-    [DataRow(nameof(RequirementTypeDeleteEvent.Plant), TestPlant)]
-    [DataRow(nameof(RequirementTypeDeleteEvent.ProjectName), null)]
-    [DataRow(nameof(RequirementTypeDeleteEvent.Behavior), "delete")]
-    public void CreateEvent_ShouldCreateRequirementDefinitionEventExpectedValues(string property, object expected)
+    [DataRow(nameof(FieldDeleteEvent.Plant), TestPlant)]
+    [DataRow(nameof(FieldDeleteEvent.ProjectName), null)]
+    [DataRow(nameof(FieldDeleteEvent.Behavior), "delete")]
+    public void CreateEvent_ShouldCreateFieldDeleteEventExpectedValues(string property, object expected)
     {
```
trondlindanger, line 27: Testing multiple behaviors within a single test method might make it harder to identify which specific case failed if the test fails. Better to split into separate test methods, each testing a single property, to adhere to principle of single state behavior?

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateTagDeleteEventHelper.cs
jsoi-eq: Coding style ought in general to be consistent througout the solution, but when new styles emerge there will be a discussion whether should we use them and whether we should change existing code to conform to new style.

@trondlindanger Do you have any opinions on this case with primary constructors, should we use them going forward, is it optional, mix and match?

richard1927:
> Btw, there seems to be test failures, so this would need to be fixed I guess before merging.

DevOps just timed out. I reran the pipeline and it came out green :)

src/Equinor.ProCoSys.Preservation.Command/EventHandlers/IntegrationEvents/EventHelpers/CreateTagDeleteEventHelper.cs
richard1927: When I think about it, both constructors have their place. Primary constructors are nice when there is no logic in the Contractors, e.g. like here where it just receives dependency injections.

However, if we need to performs some logic, like when getting a value from a configuration, then we need a "old-style" constructor. 

I think we should use both, depending on the need :)

src/tests/Equinor.ProCoSys.Preservation.Command.Tests/EventHandlers/IntegrationEvents/CreateFieldDeletedEventHelperTests.cs
richard1927: I am not familiar with the single state behaviour principle. What set of principles is that from?
I will ask you about that offline.

As to the concern to failures being harder to identify, I cant see that being a problem here. E.g. here I have changed "delete" to "delete2" to get a failure, and I think it is pretty obvious where the issue is.
image

I can see the merit of separating each of these out to their own individual test case. Then you get a clear error with a specific test case implementation that a failing assertion will point to. Here a failing assertion can refer to a number of tests.

The benefit this implementation is ease of use, and maintenance. Testing a new parameter only needs a new data row attribute. Also, if something changes in the event that changes* how we should test the properties, then we need to change fewer corresponding tests.

Personally I mostly like this behaviour since it prevents duplicate code. 

_*This should not happen (much) if we follow the open-closed principle_

trondlindanger (approved):

PR #998, opened 2025-01-28
Path fix

richard1927:
Casing in test solution paths configurations did not have correct casing. This caused a pipeline failure on Linux.

https://github.com/equinor/cc-toolbox/issues/1064


jsoi-eq (approved):
Ah, the differences between Windows and Linux...

PR #1001, opened 2025-01-30
Update to top level statements

richard1927:
Updating the application builder to use top level statements, and removing the outdated Startup class.

Changes were made using Completion API as reference.

https://github.com/equinor/cc-toolbox/issues/1064


src/Equinor.ProCoSys.Preservation.WebApi/DiModules/ServiceBusConfig.cs
```diff
@@ -0,0 +1,64 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
+
+namespace Equinor.ProCoSys.Preservation.WebApi.DiModules;
+
+public static class ServiceBusConfig
+{
+    public static void ConfigureServiceBus(this WebApplicationBuilder builder)
+    {
+        if (!builder.IsServiceBusEnabled())
```
jsoi-eq, line 14: Should we write some unit tests for this method, I'm a bit undecided myself, perhaps the code is obvious, but we have a very strong test suite in Preservation testing a lot of scenarios.
I'm thinking about testing the various branches in the if statements.

src/Equinor.ProCoSys.Preservation.WebApi/Program.cs
```diff
@@ -1,54 +1,97 @@
...
-                    {
-                        config.AddAzureAppConfiguration(options =>
-                        {
-                            var connectionString = settings["ConnectionStrings:AppConfig"];
-                            options.Connect(connectionString)
-                                .ConfigureKeyVault(kv =>
-                                {
-                                    kv.SetCredential(new ManagedIdentityCredential());
-                                })
-                                .Select(KeyFilter.Any)
```
jsoi-eq, line 34: Curious, where did the KeyFilter and Sentinel code go?

src/tests/Equinor.ProCoSys.Preservation.WebApi.IntegrationTests/TestFactory.cs
```diff
@@ -211,26 +209,12 @@ private void SeedDataForPlant(PreservationContext dbContext, IServiceProvider sc
...
+                using var sp = services.BuildServiceProvider();
+                using var dbContext = sp.GetRequiredService<PreservationContext>();
+                
+                dbContext.Database.EnsureDeleted();
             });
 
-        private PreservationContext DatabaseContext(IServiceCollection services)
-        {
-            services.AddDbContext<PreservationContext>(options 
-                => options.UseSqlServer(_connectionString, o => o.UseQuerySplittingBehavior(QuerySplittingBehavior.SplitQuery)));
```
jsoi-eq, line 223: Is all the extra code in this method unnecessary, we only need the GetRequiredService<PreservationContext> part?
Do we get the UseQuerySplittingBehavior when using the new code?

jsoi-eq (commented):
Waiting for feedback on comments.

src/Equinor.ProCoSys.Preservation.WebApi/DiModules/ServiceBusConfig.cs
richard1927: This method should have some unit tests, but we should. This is behaviour that we want validated.

The logic was originally an one line ternary operation that I migrated into a function for readability. That is one of the reasons I dislike ternary operator. It is often a way to hide logic that should be in a function that is unit tested.

I will complete that migration and add some unit tests to validate the behaviour.

src/tests/Equinor.ProCoSys.Preservation.WebApi.IntegrationTests/TestFactory.cs
richard1927: It looks like the UseQuerySplittingBehavior is not needed. all tests are passing. This was causing problems with the test where a DB context was being added to services that had already been build, causing an error.

This change was directly taken from Completion API, that does not use the DatabaseContext function either.

src/Equinor.ProCoSys.Preservation.WebApi/Program.cs
richard1927: That is a mistake on my part. It should be in the change. I will add it.

PR #1005, opened 2025-02-17
1065 service bus authentication

richard1927:
Using token credential to authenticate service bus connection

Bus sender authentication change: https://github.com/equinor/procosys-bus-sender-worker-service/pull/167

https://github.com/equinor/cc-toolbox/issues/1065

### DIFF
--- a/src/Equinor.ProCoSys.Preservation.WebApi/DiModules/ApplicationModule.cs
+++ b/src/Equinor.ProCoSys.Preservation.WebApi/DiModules/ApplicationModule.cs
@@ -1,4 +1,6 @@
-﻿using System.Text.Json.Serialization;
+﻿using System;
+using System.Text.Json.Serialization;
+using Azure.Core;
 using Equinor.ProCoSys.Auth.Authentication;
 using Equinor.ProCoSys.Auth.Authorization;
 using Equinor.ProCoSys.Auth.Caches;
@@ -60,12 +62,13 @@ using MediatR;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Action = Equinor.ProCoSys.Preservation.Domain.AggregateModels.ProjectAggregate.Action;
 
 namespace Equinor.ProCoSys.Preservation.WebApi.DIModules
 {
     public static class ApplicationModule
     {
-        public static void AddApplicationModules(this IServiceCollection services, IConfiguration configuration)
+        public static void AddApplicationModules(this IServiceCollection services, IConfiguration configuration, TokenCredential credential)
         {
             services.Configure<MainApiOptions>(configuration.GetSection("MainApi"));
             services.Configure<MainApiAuthenticatorOptions>(configuration.GetSection("MainApi"));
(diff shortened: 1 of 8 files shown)


src/Equinor.ProCoSys.Preservation.WebApi/DiModules/ApplicationModule.cs
```diff
@@ -93,8 +96,13 @@ public static void AddApplicationModules(this IServiceCollection services, IConf
 
                 x.UsingAzureServiceBus((context, cfg) =>
                 {
-                    var connectionString = configuration.GetConnectionString("ServiceBus");
-                    cfg.Host(connectionString);
+                    var serviceBusNamespace = configuration.GetConfig<string>("ServiceBus:Namespace");
```
jsoi-eq, line 99: Might consider using Options pattern.

src/Equinor.ProCoSys.Preservation.WebApi/DiModules/ServiceBusConfig.cs
```diff
@@ -8,18 +9,17 @@ namespace Equinor.ProCoSys.Preservation.WebApi.DiModules;
...
 {
-    public static void ConfigureServiceBus(this WebApplicationBuilder builder)
+    public static void ConfigureServiceBus(this WebApplicationBuilder builder, TokenCredential credential)
     {
         if (!builder.IsServiceBusEnabled())
         {
             return;
         }
+
+        var serviceBusNamespace = builder.GetConfig<string>("ServiceBus:Namespace");
```
jsoi-eq, line 19: Possible options pattern candidate.

src/Equinor.ProCoSys.Preservation.WebApi/Misc/GetConfig.cs
```diff
@@ -6,9 +6,11 @@ namespace Equinor.ProCoSys.Preservation.WebApi.Misc;
 
 public static class GetConfigExtension
 {
-    public static T GetConfig<T>(this WebApplicationBuilder builder, string configKey)
+    public static T GetConfig<T>(this WebApplicationBuilder builder, string configKey) => builder.Configuration.GetValue<T>(configKey);
```
jsoi-eq, line 9: No null check for this method, similar to the null check in method below?

richard1927: Good catch, that should be a GetConfig pass through, not GetValue

src/Equinor.ProCoSys.Preservation.WebApi/DiModules/ApplicationModule.cs
richard1927: Good idea. I will make the change.

TordJoranger: I dont see the need? We're not using it after this for anything

src/Equinor.ProCoSys.Preservation.WebApi/Misc/GetConfig.cs
TordJoranger, line 9: I rather we dont use the misc GetConfig at all, if we're just using a passthrough. 
This seems like code just to code.

how is 
builder.GetConfig<string>("ServiceBus:Namespace");
better than
builder.Configuration.GetValue<T>("ServiceBus:Namespace");
?

richard1927: GetValue returns a default value if the configuration is not found. This explicitly throws if it is not able to retrieve the configuration. This is simply here to reduce the need to check for null where we expect a configuration to be in place.

src/Equinor.ProCoSys.Preservation.WebApi/DiModules/ApplicationModule.cs
richard1927: Fair point. This configuration is just used directly during startup, not via dependency injection later down the line.

PR #1016, opened 2025-02-28
1271 dead letter message fix

richard1927:
The BusReceiverService was using user authentication tokens to authenticate main API requests. Changing to application authentication. 

https://github.com/equinor/cc-toolbox/issues/1272


janid1967 (approved):

richard1927:
> What did you do with DL's produced while the new version was active?

Once once the I had rolled back to the previous version I re-queued the messages. We only had certificate messages.

PR #1018, opened 2025-03-05
Removing redundant app settings

richard1927:
These configurations should have been removed in PR #1014

They have been moved into the application section, and the configurations default to false. Making it redundant to have them in the configuration.

https://github.com/equinor/cc-toolbox/issues/1098

### DIFF
--- a/src/Equinor.ProCoSys.Preservation.WebApi/appsettings.json
+++ b/src/Equinor.ProCoSys.Preservation.WebApi/appsettings.json
@@ -1,7 +1,9 @@
 {
   "AllowedHosts": "*",
   "Application": {
-    "DevOnLocalhost": false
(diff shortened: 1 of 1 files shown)


src/Equinor.ProCoSys.Preservation.WebApi/appsettings.json
```diff
@@ -54,8 +54,6 @@
   "MainApi": {
     "ApiVersion": "4.1"
   },
-  "MigrateDatabase": false,
```
janid1967, line 57: I prefer to move the settings inside Application section. Then the developer can read appsettings.json to find which configs can be tweaked or not

janid1967 (approved):
Approved with a suggestion

PR #1025, opened 2025-04-11
ClientCascade Project Tag OnDelete NoAction with ClientCascade 

richard1927:
Moving a tag from one project to another existing project was causing an exception in the audit data setter logic (preservation context), when accessing the change tracker.

The error would originate when the tag is removed from a project before being added to another. The entity framework would interpret this as the Tag having been set to null, even through it was added to another project.
Accessing the change tracker would then in turn throw an exception because the tag project cannot be null.

This is likely a bug in EF. I will be following up on that in the associated issue. https://github.com/equinor/cc-toolbox/issues/1642

Replacing the NoAction configuration with ClientCascade fixed the behaviour, and EF no longer attempted to set the project to null.

### DIFF
--- a/src/Equinor.ProCoSys.Preservation.Infrastructure/Entityconfigurations/ProjectConfiguration.cs
+++ b/src/Equinor.ProCoSys.Preservation.Infrastructure/Entityconfigurations/ProjectConfiguration.cs
@@ -25,8 +25,7 @@ namespace Equinor.ProCoSys.Preservation.Infrastructure.EntityConfigurations
             builder
                 .HasMany(x => x.Tags)
                 .WithOne()
-                .IsRequired()
-                .OnDelete(DeleteBehavior.NoAction);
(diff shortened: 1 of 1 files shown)


src/Equinor.ProCoSys.Preservation.Infrastructure/Entityconfigurations/ProjectConfiguration.cs
```diff
@@ -25,8 +25,7 @@ public void Configure(EntityTypeBuilder<Project> builder)
             builder
                 .HasMany(x => x.Tags)
                 .WithOne()
-                .IsRequired()
-                .OnDelete(DeleteBehavior.NoAction);
```
TordJoranger, line 29: This will add a cascading change to the DB.
I suspect ClientCascade is what we are looking for

TordJoranger (approved):

